PR #653, opened 2023-01-13
RDUCP 219 - Implement Post Message and Preview styling

os-jsplopes:
This PR is for implementing the PostMessage Capability on OutsystemsUI to handle the ODC preview application use case.

### What was happening

- N/A

### What was done

- Implemented the PostMessage listener to the window;
- Handle the message that comes from the ODC preview to add the styling for the preview devices;

### Test Steps

1. Open the app preview;
2. Check if the default device is presenting the content correctly with the notch space;
3. Change the device type;
4. Check that the previewed app notch space is being shown correctly;

### Screenshots

(prefer animated gif)

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/Event/WindowOnMessage.ts
+++ b/src/scripts/OSFramework/Event/WindowOnMessage.ts
@@ -10,13 +10,9 @@ namespace OSFramework.Event {
 	export class WindowMessage extends Event.AbstractEvent<string> {
 		constructor() {
 			super();
-			if (this._windowIsInsideIframe()) {
+			if (window.self !== window.top) {
 				window.addEventListener(GlobalEnum.HTMLEvent.Message, this._windowTrigger.bind(this), true);
 			}
-		}
-
-		private _windowIsInsideIframe(): boolean {
-			return window.self !== window.top;
 		}
 
 		private _windowTrigger(evt: MessageEvent): void {
--- a/src/scripts/OSFramework/Event/WindowOnMessage.ts
+++ b/src/scripts/OSFramework/Event/WindowOnMessage.ts
@@ -1,0 +1,26 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+namespace OSFramework.Event {
+	/**
+	 * Class that represents the Post Message on the Window.
+	 *
+	 * @export
+	 * @class WindowMessage
+	 * @extends {Event.AbstractEvent<string>}
+	 */
+	export class WindowMessage extends Event.AbstractEvent<string> {
+		constructor() {
+			super();
+			if (this._windowIsInsideIframe()) {
+				window.addEventListener(GlobalEnum.HTMLEvent.Message, this._windowTrigger.bind(this), true);
+			}
+		}
+
+		private _windowIsInsideIframe(): boolean {
+			return window.self !== window.top;
+		}
+
+		private _windowTrigger(evt: MessageEvent): void {
+			this.trigger(Event.Type.WindowMessage, evt);
+		}
+	}
+}
--- a/src/scripts/OSFramework/Helper/Device.ts
+++ b/src/scripts/OSFramework/Helper/Device.ts
@@ -130,8 +130,8 @@ namespace OSFramework.Helper {
 			const cleanedUserAgent = userAgent.replace(' ', '');
 
 			if (cleanedUserAgent === '') {
-				if (sessionStorage.userAgent) {
-					return sessionStorage.userAgent.toLowerCase();
+				if (sessionStorage.previewDevicesUserAgent) {
+					return sessionStorage.previewDevicesUserAgent.toLowerCase();
 				}
 				return window.navigator.userAgent.toLowerCase();
 			} else {
@@ -342,7 +342,10 @@ namespace OSFramework.Helper {
 		public static get IsIphoneWithNotch(): boolean {
 			if (DeviceInfo._isIphoneWithNotch === undefined) {
 				// get the device pixel ratio
-				const ratio = (sessionStorage.pixelRatio ? sessionStorage.pixelRatio : window.devicePixelRatio) || 1;
+				const ratio =
+					(sessionStorage.previewDevicesPixelRatio
+						? sessionStorage.previewDevicesPixelRatio
+						: window.devicePixelRatio) || 1;
 				const currScreen: iphoneDetails = {
 					width: window.visualViewport.width * ratio,
 					height: window.visualViewport.height * ratio,
--- a/src/scripts/OSFramework/Helper/Device.ts
+++ b/src/scripts/OSFramework/Helper/Device.ts
@@ -127,9 +127,16 @@ namespace OSFramework.Helper {
 		 * @memberof DeviceInfo
 		 */
 		private static _getUserAgent(userAgent = ''): string {
-			return userAgent.replace(' ', '') === ''
-				? window.navigator.userAgent.toLowerCase()
-				: userAgent.toLowerCase();
+			const cleanedUserAgent = userAgent.replace(' ', '');
+
+			if (cleanedUserAgent === '') {
+				if (sessionStorage.userAgent) {
+					return sessionStorage.userAgent.toLowerCase();
+				}
+				return window.navigator.userAgent.toLowerCase();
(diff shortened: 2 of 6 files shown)


src/scripts/OSFramework/Event/WindowOnMessage.ts
```diff
@@ -0,0 +1,26 @@
...
+		constructor() {
+			super();
+			if (this._windowIsInsideIframe()) {
+				window.addEventListener(GlobalEnum.HTMLEvent.Message, this._windowTrigger.bind(this), true);
+			}
+		}
+
+		private _windowIsInsideIframe(): boolean {
+			return window.self !== window.top;
+		}
```
rugoncalves, line 20: I would remove this a move the condition directly to the `if`.

src/scripts/OSFramework/Helper/Device.ts
```diff
@@ -127,9 +127,16 @@ namespace OSFramework.Helper {
...
+			const cleanedUserAgent = userAgent.replace(' ', '');
+
+			if (cleanedUserAgent === '') {
+				if (sessionStorage.userAgent) {
+					return sessionStorage.userAgent.toLowerCase();
+				}
+				return window.navigator.userAgent.toLowerCase();
+			} else {
+				return userAgent.toLowerCase();
+			}
```
rugoncalves, line 139: I don't like the approach of spreading `return` in the code. Maybe some refactoring in the future would be good.

src/scripts/OSFramework/Helper/Device.ts
```diff
@@ -127,9 +127,16 @@ namespace OSFramework.Helper {
...
 		 */
 		private static _getUserAgent(userAgent = ''): string {
-			return userAgent.replace(' ', '') === ''
-				? window.navigator.userAgent.toLowerCase()
-				: userAgent.toLowerCase();
+			const cleanedUserAgent = userAgent.replace(' ', '');
+
+			if (cleanedUserAgent === '') {
+				if (sessionStorage.userAgent) {
+					return sessionStorage.userAgent.toLowerCase();
```
rugoncalves, line 134: The name of the variable in the session storage, should be more... explicit. Something like `sessionStorage.previewDevicesUserAgent`...

src/scripts/OSFramework/Helper/Device.ts
```diff
@@ -335,10 +342,10 @@ namespace OSFramework.Helper {
 		public static get IsIphoneWithNotch(): boolean {
 			if (DeviceInfo._isIphoneWithNotch === undefined) {
 				// get the device pixel ratio
-				const ratio = window.devicePixelRatio || 1;
+				const ratio = (sessionStorage.pixelRatio ? sessionStorage.pixelRatio : window.devicePixelRatio) || 1;
```
rugoncalves, line 345: The same here: `sessionStorage.previewDevicesPixelRatio`

src/scripts/OSFramework/Helper/Device.ts
```diff
@@ -335,10 +342,10 @@ namespace OSFramework.Helper {
...
 				const currScreen: iphoneDetails = {
-					width: window.screen.width * ratio,
-					height: window.screen.height * ratio,
+					width: window.visualViewport.width * ratio,
+					height: window.visualViewport.height * ratio,
```
rugoncalves, line 348: Why was this change?

src/scripts/OutSystems/OSUI/Utils/LayoutPrivateOnPostMessage.ts
```diff
@@ -0,0 +1,104 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+namespace OutSystems.OSUI.Utils.LayoutPrivate {
```
rugoncalves, line 2: Why is this in the LayoutPrivate?

src/scripts/OutSystems/OSUI/Utils/LayoutPrivateOnPostMessage.ts
```diff
@@ -0,0 +1,104 @@
...
+				.is-phone .active-screen.screen-container::-webkit-scrollbar, html::-webkit-scrollbar,
+				.is-phone.ios .active-screen.screen-container .content::-webkit-scrollbar, html::-webkit-scrollbar {
+					display: none;
+				}
+			`;
+
+			// Adds the is phone class to apply the correct styling
+			// because platform is replacing phone with tablet when we rotate the phone device
+			document.body.classList.add('is-phone');
+			document.body.setAttribute('data-status-bar-height', `${notchValue}px`);
```
rugoncalves, line 38: ui... Why this attribute and not the variable?

src/scripts/OutSystems/OSUI/Utils/LayoutPrivateOnPostMessage.ts
```diff
@@ -0,0 +1,104 @@
...
+		 * Function used to set the orientation event
+		 */
+		public static Set(): void {
+			if (window.self !== window.top) {
+				OSFramework.Event.GlobalEventManager.Instance.addHandler(
+					OSFramework.Event.Type.WindowMessage,
+					OnPostMessage._message
+				);
+			}
+		}
```
rugoncalves, line 90: Isn't this the equivalent of making a `static constructor()`? with the advantage that it will be invoked by itself?

rugoncalves (changes requested):

src/scripts/OSFramework/Event/WindowOnMessage.ts
os-jsplopes: Moved

src/scripts/OSFramework/Helper/Device.ts
os-jsplopes: Yup the code needs some refactoring.

src/scripts/OSFramework/Helper/Device.ts
os-jsplopes: Done

src/scripts/OSFramework/Helper/Device.ts
os-jsplopes: Done

src/scripts/OSFramework/Helper/Device.ts
os-jsplopes: This was one of the errors in the console, aligned with Bernardo Cardoso about this fix. Because the screen was null at the calculation time.

src/scripts/OutSystems/OSUI/Utils/LayoutPrivateOnPostMessage.ts
os-jsplopes: The values is passed in the post message in the future if we need to adjust the size we don't need to release another OSUI version.

src/scripts/OutSystems/OSUI/Utils/LayoutPrivateOnPostMessage.ts
os-jsplopes: Since this affects the layout and we also have other methods that are adjusting the document and body and placed in the LayoutPrivate I believe this is the correct place to add it. Do you have any other suggestions?

rugoncalves: As you talked with O11 team, leave it as it is then.

src/scripts/OutSystems/OSUI/Utils/LayoutPrivateOnPostMessage.ts
rugoncalves: I missed the fact that the variable is being set by css! 🥂

src/scripts/OutSystems/OSUI/Utils/LayoutPrivateOnPostMessage.ts
rugoncalves: `static constructor()` after all is not possible in TS. Leave it as it is, because the options are too complex...

rugoncalves (approved):

bruno-colaco-os (approved):

PR #672, opened 2023-02-07
ROU-4119: fix bar border-radius with high heights values

BenOsodrac:
This PR is for fixing progressBar progress radius when the Height value is considerable high.

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Progress/Bar/scss/_progressbar.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/Progress/Bar/scss/_progressbar.scss
@@ -50,6 +50,7 @@
 
 			&:after,
 			&:before {
+				border-radius: calc(var(--shape) / 2);
 				content: '';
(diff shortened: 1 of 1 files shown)


JoaoFerreira-FrontEnd (approved):

JoaoFerreira-FrontEnd (changes requested):
This doesn't solve the problem... only reduces the chances by half.

BenOsodrac:
> This doesn't solve the problem... only reduces the chances by half.

True, but I think it's not possible to make better than this, due to the way the pattern is made. 
The value should have been an element inside the bar, not a sibling pseudo-element.

Do you have any idea?

BenOsodrac:
> This doesn't solve the problem... only reduces the chances by half.

Added new commit with your suggestion 💪

JoaoFerreira-FrontEnd (approved):

joselrio (approved):

PR #707, opened 2023-03-23
ROU-4212: Increase area of dropping elements on tabs

bmarcelino-fe:
This PR is for increase the area of dropping elements on tabs content. Please check the JIRA issue for more information

### What was happening
- Now the area to drag and drop elements is to short;

### What was done
- The area of dropping elements on content tabs was increase only when is empty;

### Screenshots
Previous:
image

After:
image

### Checklist
-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Tabs/scss/_tabs.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/Tabs/scss/_tabs.scss
@@ -234,6 +234,10 @@
 				-servicestudio-margin-bottom: var(--space-m);
 			}
 
+			& div:empty {
+				-servicestudio-padding: var(--space-m);
+			}
+
 			// This selector is used to prevent render issues on chrome, in some screen-sizes, due to overflow hidden and css snap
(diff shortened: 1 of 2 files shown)


src/scripts/OSFramework/OSUI/Pattern/Tabs/scss/_tabs.scss
```diff
@@ -234,6 +234,10 @@
 				-servicestudio-margin-bottom: var(--space-m);
 			}
 
+			& div:empty {
```
BenOsodrac, line 237: This should be inside the selector on line 232

BenOsodrac (changes requested):

src/scripts/OSFramework/OSUI/Pattern/Tabs/scss/_tabs.scss
bmarcelino-fe: Fixed

BenOsodrac (approved):

PR #712, opened 2023-03-29
RDUCP-472: Adding Provider Configs apply event

rugoncalves:
This PR is for adding the provider configs event. This event will be triggered after the setProviderConfigs methods are invoked and the new configs applied.

### What was happening
- The lack of this event, caused somethings for a flickering to occur, since the pattern is first built, and then destroyed once the method setProviderConfigs is invoked.

### What was done
- The previously abstract method `registerCallback` in the class `AbstractProviderPattern` was not implemented.
- This method encapsulates the logic of the events `Initialized` (previously dispersed through out the code) and the newly created event `OnProviderConfigsApplied`.
- Two new methods were added in the class `AbstractProviderPattern`:
  - triggerPlatformEventInitialized: will raise the `Initialized` event of the pattern;
  - triggerPlatformEventplatformCallback: will raise any generic event to OutSystems platform. This means that as first parameter there will always be the `this.widgetId`.
- Added correct context to some of the async calls

### Checklist
-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/GlobalEnum.ts
+++ b/src/scripts/OSFramework/GlobalEnum.ts
@@ -352,4 +352,15 @@ namespace OSFramework.GlobalEnum {
 	export enum NullValues {
 		Time = '00:00:00',
 	}
+
+	/**
+	 * Events available for all provider based patterns.
+	 *
+	 * @export
+	 * @enum {number}
+	 */
+	export enum ProviderEvents {
+		Initialized = 'Initialized',
+		OnProviderConfigsApplied = 'OnProviderConfigsApplied',
+	}
 }
--- a/gulp/templates/PatternFramework_HasProvider/Provider/Provider.ts
+++ b/gulp/templates/PatternFramework_HasProvider/Provider/Provider.ts
@@ -51,6 +51,7 @@ namespace Providers.<%= patternNamePC %>.<%= providerNamePC %> {
 		 */
 		protected unsetCallbacks(): void {
 			// TODO (by CreateNewPattern): Update or Remove
+			super.unsetCallbacks();
 		}
 
 		public build(): void {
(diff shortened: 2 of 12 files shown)


src/scripts/OSFramework/GlobalEnum.ts
```diff
@@ -352,4 +352,15 @@ namespace OSFramework.GlobalEnum {
 	export enum NullValues {
 		Time = '00:00:00',
```
gnbm, line 353: @rugoncalves Just a doubt. Shouldn't this be a constant instead of an enumerator?

gnbm:
@rugoncalves we should take a look at the findings from SonarCloud analysis

os-jsplopes (approved):

src/scripts/OSFramework/GlobalEnum.ts
rugoncalves: I haven't changed this. However, I think it was made to then have other null values, such as nulldate and others. Good question why that isn't there yet?!..

rugoncalves:
@gnbm the sonar cloud is complaining about duplication of code in the class `OSUIFlatpickrTime` and `OSUIFlatpickrMonth`. Any thoughts?

gnbm:
> @gnbm the sonar cloud is complaining about duplication of code in the class `OSUIFlatpickrTime` and `OSUIFlatpickrMonth`. Any thoughts?

Would need to take a look at the code. Anyway after we have the API abstraction I would like to start looking at sonarcloud in more detail

src/scripts/OSFramework/GlobalEnum.ts
gnbm: @joselrio any insight on this?

PR #713, opened 2023-03-29
RDUCP-472o11: Adding Provider Configs apply event

rugoncalves:
This PR is for adding the provider configs event. This event will be triggered after the setProviderConfigs methods are invoked and the new configs applied.

### What was happening
- The lack of this event, caused somethings for a flickering to occur, since the pattern is first built, and then destroyed once the method setProviderConfigs is invoked.

### What was done
- The previously abstract method `registerCallback` in the class `AbstractProviderPattern` was not implemented.
- This method encapsulates the logic of the events `Initialized` (previously dispersed through out the code) and the newly created event `OnProviderConfigsApplied`.
- Two new methods were added in the class `AbstractProviderPattern`:
  - triggerPlatformEventInitialized: will raise the `Initialized` event of the pattern;
  - triggerPlatformEventplatformCallback: will raise any generic event to OutSystems platform. This means that as first parameter there will always be the `this.widgetId`.
- Added correct context to some of the async calls

### Checklist
-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/Providers/OSUI/RangeSlider/NoUISlider/AbstractNoUiSlider.ts
+++ b/src/scripts/Providers/OSUI/RangeSlider/NoUISlider/AbstractNoUiSlider.ts
@@ -170,6 +170,7 @@ namespace Providers.OSUI.RangeSlider.NoUISlider {
 		 */
 		protected unsetCallbacks(): void {
 			this.eventProviderValueChanged = undefined;
+			super.unsetCallbacks();
 		}
 
 		/**
@@ -312,6 +313,7 @@ namespace Providers.OSUI.RangeSlider.NoUISlider {
 			this.configs.setExtensibilityConfigs(newConfigs);
 
 			this.redraw();
+			super.setProviderConfigs(newConfigs);
 		}
 
 		/**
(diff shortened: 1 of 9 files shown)


src/scripts/Providers/OSUI/RangeSlider/NoUISlider/AbstractNoUiSlider.ts
```diff
@@ -319,6 +313,7 @@ namespace Providers.OSUI.RangeSlider.NoUISlider {
 			this.configs.setExtensibilityConfigs(newConfigs);
 
 			this.redraw();
+			super.setProviderConfigs(newConfigs);
```
joselrio, line 316: Shouldn't this be before the redraw?

src/scripts/Providers/OSUI/Timepicker/Flatpickr/FlatpickrTime.ts
```diff
@@ -381,6 +376,8 @@ namespace Providers.OSUI.TimePicker.Flatpickr {
 			this.configs.setExtensibilityConfigs(newConfigs);
 
 			this.redraw();
+
+			super.setProviderConfigs(newConfigs);
```
joselrio, line 380: Same as before.

BenOsodrac (approved):

PR #724, opened 2023-04-17
ROU-3789: Fix issue on layout native

bmarcelino-fe:
This PR is for fixing an issue on layout native when the menu behavior is expandable.

### What was happening
- The CSS rules are incorrectly applied:
image
image

### What was done
- Removed the CSS that are causing the issue:
image
image

### Test Steps
1. Open Sample Page
2. Check if the layout are correct
3. Enable the RTL
4. Expected: The layout will change the view based on RTL and the menu icon are visible

### Checklist
-   [X] tested locally
-   [X] documented the code


dist/OutSystemsUI.d.ts
```diff
@@ -2958,8 +2959,7 @@ declare namespace OSFramework.OSUI.Patterns.Tabs {
         private _hasDragGestures;
         private _hasSingleContent;
         private _headerItemsLength;
-        private _isChrome;
-        private _isEdge;
```
gnbm, line 2962: @BenOsodrac Wasn't this needed and different from the _isChromium?

gnbm: Forget it..the content of the dit shouldn't be in the PR I would say

BenOsodrac (approved):

PR #726, opened 2023-04-18
ROU-4203: Change the moved elements to active screen

bmarcelino-fe:
This PR is for change the target element of moved ballon from out of context of the pattern. The Tooltip and the DropdownServerSide were affected with this change.

### Checklist
-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/DropdownServerSide.ts
+++ b/src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/DropdownServerSide.ts
@@ -63,8 +63,6 @@ namespace OSFramework.OSUI.Patterns.Dropdown.ServerSide {
 		private _isBlocked = false;
 		// Store the Element State, by default is closed!
 		private _isOpen = false;
-		// Store the HTML element for the layout where the Balloon will be moved into.
-		private _layoutElement: HTMLElement;
 		// Platform OnInitialize Callback
 		private _platformEventInitializedCallback: GlobalCallbacks.OSGeneric;
 		// Platform OnClose Callback
@@ -111,7 +109,7 @@ namespace OSFramework.OSUI.Patterns.Dropdown.ServerSide {
 			}
 
 			// Remove isVisible class to the layout
-			Helper.Dom.Styles.RemoveClass(this._activeScreenElement, Enum.CssClass.IsVisible);
+			Helper.Dom.Styles.RemoveClass(document.body, Enum.CssClass.IsVisible);
 
 			// Update the touchMove when pattern is open!
 			this._touchMove();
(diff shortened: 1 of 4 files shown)


src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/DropdownServerSide.ts
```diff
@@ -111,7 +109,7 @@ namespace OSFramework.OSUI.Patterns.Dropdown.ServerSide {
 			}
 
 			// Remove isVisible class to the layout
-			Helper.Dom.Styles.RemoveClass(this._activeScreenElement, Enum.CssClass.IsVisible);
+			Helper.Dom.Styles.RemoveClass(document.body, Enum.CssClass.IsVisible);
```
BenOsodrac, line 112: comment:
*to the body

src/scripts/OSFramework/OSUI/Pattern/Tooltip/Tooltip.ts
```diff
@@ -59,10 +59,13 @@ namespace OSFramework.OSUI.Patterns.Tooltip {
 			this._tooltipBalloonPositionClass = this.configs.Position;
 		}
 
-		// Move balloon element to outside of the pattern context
+		// Move balloon element to active screen, outside of the pattern context
```
BenOsodrac, line 62: comment:
*to the body

src/scripts/OSFramework/OSUI/Pattern/Tooltip/Tooltip.ts
```diff
@@ -59,10 +59,13 @@ namespace OSFramework.OSUI.Patterns.Tooltip {
...
 		private _moveBalloonElement(): void {
-			const layoutElement = Helper.Dom.ClassSelector(document.body, GlobalEnum.CssClassElements.Layout);
-			Helper.Dom.Move(this._tooltipBalloonWrapperElem, layoutElement);
+			const activeScreenElement = Helper.Dom.ClassSelector(
```
BenOsodrac, line 64: Is this method called multiple times? We should define the activeScreenElement screen only once. 
Maybe on the setHtmlElements method?

BenOsodrac (changes requested):

src/scripts/OSFramework/OSUI/Pattern/Tooltip/Tooltip.ts
bmarcelino-fe: The element is moved to active-screen and not to the body.

src/scripts/OSFramework/OSUI/Pattern/Tooltip/Tooltip.ts
bmarcelino-fe: However, to keep consistency, I've change the way that we are defining the element and do the same approach as Dropdown.

BenOsodrac (approved):

joselrio (approved):

PR #733, opened 2023-05-03
ROU-4245: Fix MaxDate validation when using DateTime

BenOsodrac:
This PR is for fixing am issue that was causing the Today Button to trigger a null Date, when the MaxDate was set to 'today'.

This was happening because inside Service Studio, the parameters are DateTime type, and when used to set Date types, the Date would come with a fixed hour ,even if not used, which could make the validations comparisons on the provider to fail, as it could happen that the Date on the Today Button would have a later hour, or even just minutes.

### What was done
- Added a new private method to check if TimeFormat is disabled. If it is, then we set the MaxDate to 23:59:59 hour and the MinDate to 0:0:0:0 hour.

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Helper/Dates.ts
+++ b/src/scripts/OSFramework/OSUI/Helper/Dates.ts
@@ -53,6 +53,33 @@ namespace OSFramework.OSUI.Helper {
 		}
 
 		/**
+		 * Funciton used to normalize the OUtSystems DateTimes, used on scopes expecting a Date
+		 *
+		 * @static
+		 * @param {(string | Date)} date
+		 * @param {boolean} [normalizeToMax=true]
+		 * @return {*}  {Date}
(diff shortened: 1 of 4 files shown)


OS-giulianasilva (approved):

src/scripts/OSFramework/OSUI/Helper/Dates.ts
```diff
@@ -52,6 +52,33 @@ namespace OSFramework.OSUI.Helper {
 			return false;
 		}
 
+		/**
+		 * Funciton used to normalize the OUtSystems DateTimes, used on scopes expecting a Date
```
joselrio, line 56: Function *

joselrio (approved):

OS-giulianasilva (approved):

PR #742, opened 2023-05-13
Patch 4

thekingofspain:
This PR is for...

* Added generic mixin to create space based attributes classes.
* Added a zero based flex option.
* Added new flew-flow, flex-grow, and flex-shrink, flex-basis classes.
* Added **display: flex** css via the a _flex-parent_ SASS template to the main flex CSS classes that are used on the flex parent element.
  * This might be considered a potential client breaking change.  Example:
  * 
     A combination of OS UI Css class are used to to set the direction or justify-content but there is some custom CSS that set the display value to inline-flex, has the potential to be overriden to the _flex_ value.
  *  Note OutSystem UI does not have a CSS class for inline-flex.  OutSystems UI does use display inline-flex, I have not cross referenced ServiceStudio code to see if there is a potential conflict.
* Added new flex-justify-content classes that inherit from _flex-parent_ SASS template.
  * Existing justify-content css classes were kept as is to preserve and allow any cross usage between _display_ values of _flex_ and _grid_.

[Sample page](url)

### What was happening

-

### What was done

-

### Test Steps

1.

### Screenshots

(prefer animated gif)

### Checklist

-   [ ] tested locally
-   [ ] documented the code

### DIFF
--- a/src/scss/00-abstract/_mixins.scss
+++ b/src/scss/00-abstract/_mixins.scss
@@ -23,3 +23,16 @@
 	position: absolute;
 	top: $top;
 }
+
+///
+@mixin osui-sized($attribute)
+{
+    /* #{$attribute} */
+    .#{$attribute} {
+	    @each $type, $value in $osui-sizes {
+	    	&-#{$type} {
+    			#{$attribute}: var($value);
+	    }
+    	}
+    }
+}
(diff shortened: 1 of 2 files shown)


joselrio:
Hi @thekingofspain,

First of all, thanks for your suggestion.
It makes sense, however we must check if this is something we want by default and run all the tests in order to avoid breaking changes as also you mentioned.
That said, we will take a look on it.

Thank you!

PR #755, opened 2023-06-01
Fixed issue that was preventing OnBodyScroll event to be triggered.

joselrio:
This PR will fix an issue related with the OnBodyScroll. 
Since the event is being set to the body element and in some cases the scroll is being set to other elements, such as ".active-screen" container, once scroll occurs inside this last one, the event was not being triggered at the body since body has an overflow hidden.
In order to grant all the use cases, **useCapture** property has been added in order to proper manage this event property accordingly.

### DIFF
--- a/src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/AbstractListener.ts
+++ b/src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/AbstractListener.ts
@@ -16,12 +16,34 @@ namespace OSFramework.OSUI.Event.DOMEvents.Listeners {
 		private _eventTarget: HTMLElement;
 		// Store the listener type
 		private _eventType: GlobalEnum.HTMLEvent;
-		// Store the listener callback
+
+		/**
+		 * Store the listener callback
+		 *
+		 * @protected
+		 * @type {EventListenerObject}
+		 * @memberof AbstractListener
+		 */
 		protected eventCallback: EventListenerObject;
 
-		constructor(eventTarget, eventType) {
+		/**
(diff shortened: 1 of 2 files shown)


src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/AbstractListener.ts
```diff
@@ -16,12 +16,34 @@ namespace OSFramework.OSUI.Event.DOMEvents.Listeners {
...
+		protected useCapture = false;
+
+		/**
+		 * Creates an instance of AbstractListener.
+		 *
+		 * @param {unknown} eventTarget, window can be in use, so, HTMLElement can't be the type of the input by default
+		 * @param {GlobalEnum.HTMLEvent} eventType
+		 * @memberof AbstractListener
+		 */
+		constructor(eventTarget: unknown, eventType: GlobalEnum.HTMLEvent) {
```
rugoncalves, line 44: Why the type `unknown`? After all, 2 lines after we're doing the "cast"...

rugoncalves (changes requested):
The comments on the attribute `useCapture` should be improved, as they are not explicit enough. ;)
Other than this, good catch!

src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/AbstractListener.ts
joselrio: Done!
Thanks for the heads up mate!

rugoncalves (approved):
👍

PR #762, opened 2023-06-19
ROU-4390: Fix InlineSVG inside progress-circle

BenOsodrac:
This PR is for fixing a breaking-change on the SVGInline when used inside the ProgressCircle, due to the CSS Class change on the SVG Pattern.

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Progress/Circle/scss/_progresscircle.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/Progress/Circle/scss/_progresscircle.scss
@@ -77,7 +77,7 @@
 		z-index: var(--layer-local-tier-1);
 	}
 
-	.svg-wrapper {
+	.osui-inline-svg {
 		display: inline-flex;
 		// the !important is used to override the inline style
 		height: var(--circle-size) !important;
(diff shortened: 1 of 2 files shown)


OS-giulianasilva (approved):

src/scripts/OSFramework/OSUI/Pattern/Progress/Circle/scss/_progresscircle.scss
```diff
@@ -77,7 +77,7 @@
 		z-index: var(--layer-local-tier-1);
 	}
 
-	.svg-wrapper {
+	.osui-inline-svg {
```
gnbm, line 80: @BenOsodrac For instance on the `osui-progress-circle` you kept svg-wrapper. Shouldn't we have both `svg-wrapper` and `osui-inline-svg`?

BenOsodrac: Don't think so, on the CSS we should use the most recent classes, no need to keep referencing old ones. The OutSystems UI is released as a bulk, so this CSS will have correct structure on the OutSystems UI module' html.

joselrio (approved):

PR #765, opened 2023-06-19
[ROU-4360] - Dropdown Search/Tags doesn't close when is open before data is loaded

joselrio:
This PR is for fixing an issue that was preventing Dropdown to work when data was completed loaded and Dropdown was opened.

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/InlineSvg/InlineSvg.ts
+++ b/src/scripts/OSFramework/OSUI/Pattern/InlineSvg/InlineSvg.ts
@@ -70,10 +70,13 @@ namespace OSFramework.OSUI.Patterns.InlineSvg {
 			console.log(GlobalEnum.WarningMessages.MethodNotImplemented);
 		}
 
+		/**
+		 * Builds the pattern.
+		 *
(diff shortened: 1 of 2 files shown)


src/scripts/OSFramework/OSUI/Pattern/InlineSvg/InlineSvg.ts
```diff
@@ -70,11 +70,14 @@ namespace OSFramework.OSUI.Patterns.InlineSvg {
 			console.log(GlobalEnum.WarningMessages.MethodNotImplemented);
```
gnbm, line 70: @joselrio Weren't these the console logs we talked about earlier?

BenOsodrac (approved):

src/scripts/OSFramework/OSUI/Pattern/InlineSvg/InlineSvg.ts
joselrio: This consoles must be there since the method is implementation mandatory, it was the call of this methods that have been removed.

OS-giulianasilva (approved):

PR #775, opened 2023-06-26
ROU-4405: Minor fixes on OverflowMenu 

BenOsodrac:
This PR is for adding the following fixes to the OverflowMenu:

- Fix shape rounded style in mobile;
- Fix hover style when shape is not Sharp and a link is the first child inside the Balloon.
- Fixed Esc behaviour inside Sidebar. Now when pressing Esc, the Sidebar is not closed, only the OverflowMenu.
- Added missing configs validations.

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/dist/OutSystemsUI.js
+++ b/dist/OutSystemsUI.js
@@ -1812,6 +1812,7 @@ var OSFramework;
                         if (isEscapedPressed && this.isOpen) {
                             this.close();
                         }
+                        e.stopPropagation();
                     }
                     _removeEventListeners() {
                         this.featureElem.removeEventListener(OSUI.GlobalEnum.HTMLEvent.keyDown, this._eventOnKeypress);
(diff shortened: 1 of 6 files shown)


dist/OutSystemsUI.js
```diff
@@ -7304,6 +7305,21 @@ var OSFramework;
...
                         super(config);
                     }
+                    validateDefault(key, value) {
+                        let validatedValue = undefined;
+                        switch (key) {
+                            case OverflowMenu.Enum.Properties.Shape:
+                                validatedValue = this.validateInRange(value, OSUI.GlobalEnum.ShapeTypes.SoftRounded, OSUI.GlobalEnum.ShapeTypes.Sharp, OSUI.GlobalEnum.ShapeTypes.Rounded);
+                                break;
+                            case OverflowMenu.Enum.Properties.Position:
+                                validatedValue = this.validateString(value, OSUI.GlobalEnum.FloatingPosition.Auto);
```
OS-giulianasilva, line 7315: Why don't we use validateInRange here like we do for the shape property?

BenOsodrac: Good point, I followed the same approach as in Notification. I will change it 👍

BenOsodrac: Fixed!

PR #781, opened 2023-07-10
ROU-4327: Improve event management on remove handler

BenOsodrac:
This PR is for improving the management of the events. This was found when improving the lifecycle of the orientationChange event (OutSystems code), that was not being added again in runtime, as the _events map wasn't correctly removing the eventType, when the last handler was removed.

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Event/DOMEvents/AbstractEvent.ts
+++ b/src/scripts/OSFramework/OSUI/Event/DOMEvents/AbstractEvent.ts
@@ -17,7 +17,7 @@ namespace OSFramework.OSUI.Event.DOMEvents {
 		 * Getter for handlers
 		 *
 		 * @readonly
-		 * @protected
+		 * @public
 		 * @type {GlobalCallbacks.OSGeneric[]}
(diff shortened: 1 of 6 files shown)


OS-giulianasilva (dismissed):

src/scripts/OSFramework/OSUI/Event/DOMEvents/AbstractEvent.ts
```diff
@@ -21,7 +21,7 @@ namespace OSFramework.OSUI.Event.DOMEvents {
 		 * @type {GlobalCallbacks.OSGeneric[]}
 		 * @memberof AbstractEvent
 		 */
-		protected get handlers(): GlobalCallbacks.OSGeneric[] {
+		public get handlers(): GlobalCallbacks.OSGeneric[] {
```
joselrio, line 24: @BenOsodrac please update comment method modifier as well. from @protected into @public

joselrio (changes requested):

joselrio (approved):

PR #794, opened 2023-07-19
Video: Change the time type to be applied on VideoJumpTime

bmarcelino-fe:
This PR is for changing the behavior of applying time in VideoJumpToTime. Now the the time is defined as time and we convert the value into seconds to apply on video.

### What was done
- Now the the time is defined as time and we convert the value into seconds to apply on video.

### Checklist
-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Video/Video.ts
+++ b/src/scripts/OSFramework/OSUI/Pattern/Video/Video.ts
@@ -13,6 +13,8 @@ namespace OSFramework.OSUI.Patterns.Video {
 		private _platformEventOnStateChanged: Callbacks.OSOnStateChangedEvent;
 		// Store the video element
 		private _videoElement: HTMLVideoElement;
+		// Store the value for video to jump into a specific time
+		private _videoJumpTime: number;
 		// Store the source element
 		private _videoSourceElement: HTMLSourceElement;
 		// Store the current video state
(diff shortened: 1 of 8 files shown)


src/scripts/OSFramework/OSUI/Pattern/Video/Video.ts
```diff
@@ -411,8 +413,14 @@ namespace OSFramework.OSUI.Patterns.Video {
 		 * @param {number} time value in seconds
 		 * @memberof Video
 		 */
-		public setVideoJumpToTime(time: number): void {
-			this._videoElement.currentTime = time;
+		public setVideoJumpToTime(time: string): void {
+			const jumpToTime = time.split(':');
```
BenOsodrac, line 417: use _ for private variables

src/scripts/OSFramework/OSUI/Pattern/Video/Video.ts
```diff
@@ -411,8 +413,14 @@ namespace OSFramework.OSUI.Patterns.Video {
...
+			// Convert the time provided into seconds
+			const timeInSeconds =
```
BenOsodrac, line 419: same as above

src/scripts/OSFramework/OSUI/Pattern/Video/Video.ts
```diff
@@ -411,8 +413,14 @@ namespace OSFramework.OSUI.Patterns.Video {
...
+				+jumpToTime[0] * Enum.VideoTime.Hour * Enum.VideoTime.Minute +
+				+jumpToTime[1] * Enum.VideoTime.Minute +
+				+jumpToTime[2];
+			this._videoElement.currentTime = timeInSeconds;
```
BenOsodrac, line 423: @bmarcelino-fe Why do we need this to be string in here? Because of the Time type the platform generates?

If that so, I would say it would be better to do that before, before calling the API. We already in other locations I think, like a Helper to translate platform time to expected js time.

That way we wouldn't pollute the javascript API

BenOsodrac (changes requested):

OS-giulianasilva (approved):

OS-giulianasilva (approved):

PR #798, opened 2023-08-03
ROU-4348: Add aria-label to the ProgressBar/Circle when empty content

OS-giulianasilva:
This PR is for add aria-label to the ProgressBar/Circle when empty content

### What was happening
- Accessibility checkers were indicating an issue when the element referenced in aria-labelledby attribute is not readable.

### What was done
- When the content placeholder contains any element, we replace the aria-label attribute from Progress Bar and Circle with the aria-labelledby referencing the Content id, otherwise, we keep the aria-label attribute.
- If the content inside the Content placeholders is not readable, it is the responsibility of the user to make it accessible.

### Test Steps
1. Go to a page with a Progress Bar or Progress Circle
2. Enable the screen reader
3. Check that the screen reader reads the Progress Bar/Circle percentage, content, and the role "progressbar".

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Progress/Bar/ProgressBar.ts
+++ b/src/scripts/OSFramework/OSUI/Pattern/Progress/Bar/ProgressBar.ts
@@ -95,7 +95,8 @@ namespace OSFramework.OSUI.Patterns.Progress.Bar {
 		 * @memberof OSFramework.Patterns.Progress.Bar.Bar
 		 */
 		protected setA11YProperties(): void {
-			Helper.A11Y.AriaLabelledBy(this.selfElement, this.contentElemId);
+			if (this.contentElem.innerHTML) Helper.A11Y.AriaLabelledBy(this.selfElement, this.contentElem.id);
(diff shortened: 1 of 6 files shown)


BenOsodrac (approved):

src/scripts/OSFramework/OSUI/Pattern/Progress/Bar/ProgressBar.ts
```diff
@@ -95,7 +95,8 @@ namespace OSFramework.OSUI.Patterns.Progress.Bar {
 		 * @memberof OSFramework.Patterns.Progress.Bar.Bar
 		 */
 		protected setA11YProperties(): void {
-			Helper.A11Y.AriaLabelledBy(this.selfElement, this.contentElemId);
+			if (this.contentElem.innerHTML) Helper.A11Y.AriaLabelledBy(this.selfElement, this.contentElem.id);
+			else Helper.A11Y.AriaLabel(this.selfElement, ProgressEnum.AriaLabel.Progress);
```
joselrio, line 99: Have mix feelings about this Progress enum due to translations, but this is something that can be changed through extensibility...

joselrio (approved):

PR #800, opened 2023-08-07
ROU-4462: Abstraction to focus on the trigger when we close a component (sidebar, notification, etc) 

OS-giulianasilva:
This PR is for create the abstraction to focus on the trigger when we close a component (sidebar, notification, balloon and bottom sheet) 

### What was happening
- When we close the notification, stay focused on the notification element. The element is hidden but the screen reader keeps focused on the element.
- When the sidebar is closed, it does not return focus to the previous focusable element
- When a component starts opening, the focus was set to the body when closed.

### What was done
- Created the FocusManager class to be instantiated in overlay components to manage the focus when closing the component.
   - The FocusManager.storeLastFocusedElement stores the element activeElement reference
   - The FocusManager.returnFocusToElement return the focus to the stored element. If the element does not exist on the page, is not focusable or is disabled, the focus is set to the first focusable element in the page.

### Test Steps
1. Go to a page with a notification that starts opening
2. Check that when the notification is closed, the focus goes to the first focusable element on the page

1. Go to a page with button to open the notification
2. Click on the button.
3. Check that when the notification is closed, the focus returns to the button.

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Behaviors/FocusManager.ts
+++ b/src/scripts/OSFramework/OSUI/Behaviors/FocusManager.ts
@@ -20,7 +20,7 @@ namespace OSFramework.OSUI.Behaviors {
 		 * @return {*}  {void}
 		 * @memberof OSFramework.Behaviors.FocusManager
 		 */
-		public returnFocusToElement(): void {
+		public setFocusToStoredElement(): void {
 			// If no element was set as the last focused element or it is not focusable anymore,
 			// return the focus to the first focusable element in the page.
 			if (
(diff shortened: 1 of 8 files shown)


src/scripts/OSFramework/OSUI/Behaviors/FocusManager.ts
```diff
@@ -0,0 +1,57 @@
...
+			// do nothing.
+		}
+
+		/**
+		 * Method to return the focus to the previous focused element
+		 *
+		 * @return {*}  {void}
+		 * @memberof OSFramework.Behaviors.FocusManager
+		 */
+		public returnFocusToElement(): void {
```
gnbm, line 23: @OS-giulianasilva Having a method name starting with "return" is a bit strange (and not even a get since it doesn't return anything). We could have something like "focusOnPreviousElement" or something better instead.

OS-giulianasilva: done!

JoaoFerreira-FrontEnd (approved):

PR #816, opened 2023-09-08
Add class

rugoncalves:
This PR is for improving the code used when the app is within the preview in devices.

### DIFF
--- a/src/scripts/OSFramework/Helper/Device.ts
+++ b/src/scripts/OSFramework/Helper/Device.ts
@@ -127,16 +127,17 @@ namespace OSFramework.Helper {
 		 * @memberof DeviceInfo
 		 */
 		private static _getUserAgent(userAgent = ''): string {
-			const cleanedUserAgent = userAgent.replace(' ', '');
+			let localUserAgent = userAgent;
 
-			if (cleanedUserAgent === '') {
+			if (userAgent.replace(' ', '') === '') {
 				if (sessionStorage.previewDevicesUserAgent) {
-					return sessionStorage.previewDevicesUserAgent.toLowerCase();
+					localUserAgent = sessionStorage.previewDevicesUserAgent;
+				} else {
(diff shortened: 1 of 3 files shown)


src/scripts/OutSystems/OSUI/Utils/PreviewInDevices/OnPostMessage.ts
```diff
@@ -1,6 +1,33 @@
...
+
+		/**
+		 * Adds a css class to the body, to make it easily style accordingly.
+		 *
+		 * @private
+		 * @static
+		 * @memberof OnPostMessage
+		 */
+		private static _addInPreviewCssClass(): void {
+			OnPostMessage._isInPreviewInDevices = true;
```
BenOsodrac, line 27: Shouldn't this be at the end of the _messageFromPreview() method?

BenOsodrac: Feels weird to set it as true, on a method that only adds a class

rugoncalves: Can be, there's no specific advantage in either of the ways.

src/scripts/OSFramework/Helper/Device.ts
```diff
@@ -127,16 +127,17 @@ namespace OSFramework.Helper {
 		 * @memberof DeviceInfo
 		 */
 		private static _getUserAgent(userAgent = ''): string {
-			const cleanedUserAgent = userAgent.replace(' ', '');
+			let localUserAgent = userAgent;
 
-			if (cleanedUserAgent === '') {
+			if (userAgent.replace(' ', '') === '') {
```
gnbm, line 132: Could also be `userAgent.replace(/\s/g, "")`?!

BenOsodrac (approved):

PR #821, opened 2023-09-12
ROU-4540: Fix issue on Datepicker that show the hidden datepicker input

bmarcelino-fe:
This PR is for fixing an issue on Datepicker that show the hidden datepicker input.

### What was happening
- The Datepicker are showing the hidden input
image

### What was done
- Change the selectors on CSS to hide the input
image

### Test Steps
1. Open test page
2. Check the Pickers input
3. Expected: no duplicated input are shown

### Checklist
-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/DatePicker/scss/_datepicker.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/DatePicker/scss/_datepicker.scss
@@ -34,14 +34,12 @@
 	}
 
 	// Hide the platform input which is set as hidden by the library and we're change it into the expected type, however we do not want it visible since library will add a clone to better deal with the selected dates.
-	& span > input.flatpickr-input {
-		&:first-of-type {
-			display: none;
+	input.flatpickr-input:first-of-type {
+		display: none;
 
-			// Make the platform input visible in Service Studio
-			& {
-				-servicestudio-display: inline-flex !important;
-			}
+		// Make the platform input visible in Service Studio
+		& {
+			-servicestudio-display: inline-flex !important;
 		}
 	}
 
--- a/src/scripts/OSFramework/OSUI/Pattern/DatePicker/scss/_datepicker.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/DatePicker/scss/_datepicker.scss
@@ -34,9 +34,9 @@
 	}
 
 	// Hide the platform input which is set as hidden by the library and we're change it into the expected type, however we do not want it visible since library will add a clone to better deal with the selected dates.
-	& > span > input {
+	& span > input.flatpickr-input {
 		&:first-of-type {
-			display: none !important;
+			display: none;
 
 			// Make the platform input visible in Service Studio
 			& {
--- a/src/scripts/OSFramework/OSUI/Pattern/MonthPicker/scss/_monthpicker.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/MonthPicker/scss/_monthpicker.scss
@@ -56,9 +56,9 @@
 	}
 
 	// Hide the platform input which is set as hidden by the library and we're change it into the expected type, however we do not want it visible since library will add a clone to better deal with the selected dates.
-	& > span > input {
+	& span > input.flatpickr-input {
 		&:first-of-type {
-			display: none !important;
+			display: none;
 
 			// Make the platform input visible in Service Studio
 			& {
--- a/src/scripts/OSFramework/OSUI/Pattern/MonthPicker/scss/_monthpicker.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/MonthPicker/scss/_monthpicker.scss
@@ -56,14 +56,12 @@
 	}
 
 	// Hide the platform input which is set as hidden by the library and we're change it into the expected type, however we do not want it visible since library will add a clone to better deal with the selected dates.
-	& span > input.flatpickr-input {
-		&:first-of-type {
-			display: none;
+	input.flatpickr-input:first-of-type {
+		display: none;
 
-			// Make the platform input visible in Service Studio
-			& {
-				-servicestudio-display: inline-flex !important;
-			}
+		// Make the platform input visible in Service Studio
+		& {
+			-servicestudio-display: inline-flex !important;
 		}
 	}
 }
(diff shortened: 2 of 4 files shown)


src/scripts/OSFramework/OSUI/Pattern/DatePicker/scss/_datepicker.scss
```diff
@@ -34,9 +34,9 @@
 	}
 
 	// Hide the platform input which is set as hidden by the library and we're change it into the expected type, however we do not want it visible since library will add a clone to better deal with the selected dates.
-	& > span > input {
+	& span > input.flatpickr-input {
```
JoaoFerreira-FrontEnd, line 37: Why we don't simplify this even more by changing this to 
".flatpickr-input:first-of-type"
?

src/scripts/OSFramework/OSUI/Pattern/MonthPicker/scss/_monthpicker.scss
```diff
@@ -56,9 +56,9 @@
 	}
 
 	// Hide the platform input which is set as hidden by the library and we're change it into the expected type, however we do not want it visible since library will add a clone to better deal with the selected dates.
-	& > span > input {
+	& span > input.flatpickr-input {
 		&:first-of-type {
```
JoaoFerreira-FrontEnd, line 60: here the same: ".flatpickr-input:first-of-type"

src/scripts/OSFramework/OSUI/Pattern/TimePicker/scss/_timepicker.scss
```diff
@@ -34,9 +34,9 @@
 	}
 
 	// Hide the platform input which is set as hidden by the library and we're change it into the expected type, however we do not want it visible since library will add a clone to better deal with the selected dates.
-	& > span > input {
+	& span > input.flatpickr-input {
 		&:first-of-type {
```
JoaoFerreira-FrontEnd, line 38: here the same: ".flatpickr-input:first-of-type"

JoaoFerreira-FrontEnd (changes requested):

src/scripts/OSFramework/OSUI/Pattern/DatePicker/scss/_datepicker.scss
bmarcelino-fe: @JoaoFerreira-FrontEnd I kept the same approach has before to guarantee a direct selector of the element, the first input with class. We never know if a user can add an extra input with this clas, in SS or by JS.

gnbm: What will we gain from adding specificity on the selector and removing the `!important`, when we never want to show this input that is cloned and added by the library? Just to try to have both perspectives since keeping the same approach shouldn't work as a justification. What are the pros and cons?
@bmarcelino-fe @JoaoFerreira-FrontEnd 
cc: @BenOsodrac

src/scripts/OSFramework/OSUI/Pattern/TimePicker/scss/_timepicker.scss
bmarcelino-fe: @JoaoFerreira-FrontEnd In this case that is what we want. We always want to keep the input hidded in all use cases. This input is hidden and is cloned for the datepicker.

src/scripts/OSFramework/OSUI/Pattern/DatePicker/scss/_datepicker.scss
bmarcelino-fe: @gnbm I can't remember the rational of using the !important here. In my perspective, if we have a solution to avoid it, it's better. With this approach and based on my tests, the input cloned is always hidden.
Related to the **:first-of-type** is to make sure that only the cloned element is affected. If the user adds the class .flatpickr-input to the input for some reason, we will hide the element from **dom**.

BenOsodrac: The first lesson of CSS is always: don't use !important, use specificity and cascase if you can.

We should just use !important when there's absolutely no other way to fix something. And here we have other ways.

Its a bad practice, that will increase the specificity of this rule to the maximum, and probably trigger warnings on CSS audit tools.

BenOsodrac: So, I agree with @bmarcelino-fe solution here.

JoaoFerreira-FrontEnd: I still don't understand why we need the selector to have the span and the element input. But if we put ourselves in a weak position by removing let's go on.

src/scripts/OSFramework/OSUI/Pattern/TimePicker/scss/_timepicker.scss
JoaoFerreira-FrontEnd: I did understand the usecase, what I've tried to share is that here my suggestion is the some as the above, but needs to be a bit more complex, because we have other rule in the file that will be stronger than ".flatpickr-input:first-of-type". that's why I've suggest... "input.flatpickr-input:first-of-type"

JoaoFerreira-FrontEnd (approved):

OS-giulianasilva (dismissed):

src/scripts/OSFramework/OSUI/Pattern/TimePicker/scss/_timepicker.scss
bmarcelino-fe: The Datepicker needs to ha 2 inputs, one from the provider and one from that platform. One of them is cloned and we need to hide the other one from DOM. This is how we have the datepicker implemented and can match the date from provider and platform. The use case is, the hidden input is being show in some use cases because the selector on OSUI is added only for the direct child. If the user adds a custom content or wrap the inpout with the container, the selector will not work.

Tha being said, we need to be more specific and remove the !important from the selector. So all the inputs with the class .flatpickr-input will be hidden and only the first one, that we controls. If the users want to add the flatpickr class for some reason, we will not hide the element.

JoaoFerreira-FrontEnd (approved):

OS-giulianasilva (approved):

PR #840, opened 2023-10-09
ROU-4570: Datepicker/DatepickerRange doesn't get focused when is opened via API 

bmarcelino-fe:
This PR is to fix issue on Datepicker Open API to focus on input element when is triggered

### What was happening
- When the Datepicker is opened using the API, the focus are the button clicked
Screenshot 2023-10-02 at 11 36 01

### What was done
- The Datepicker open API was reviewed to focus the input element when is triggered, to keep the same behavior as the default when is opened by input click
image

### Test Steps

1. Go to each sample page
2. Enable the A11Y switch
3. Click on apply button
4. Open the datepicker by the button on right top of input
5. Expected: the datepicker will open and the input will be focused

### Checklist
-   [X] tested locally
-   [X] documented the code


src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
```diff
@@ -495,6 +495,9 @@ namespace Providers.OSUI.Datepicker.Flatpickr {
 			const isInputDisable = this.datePickerPlatformInputElem.disabled;
 			if (this.provider.isOpen === false && isInputDisable === false) {
 				this.provider.open();
+
+				// Focus on input element, after the open trigger occur
+				this.flatpickrInputElem.focus();
```
BenOsodrac, line 500: I would change to `this.flatpickrInputElem?.focus()` to validate first if the element exist. We can be using the fixed option or the native one, without the input

BenOsodrac (approved):

BenOsodrac (approved):

joselrio (approved):

PR #846, opened 2023-10-23
Update _dropdown.scss

alexandre97costa:
Added styling to the dropdown border when in focused state.

This PR is for Dropdown Search styling

### What was happening

-

### What was done

-

### Test Steps

1.

### Screenshots

(prefer animated gif)

### Checklist

-   [ ] tested locally
-   [ ] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Dropdown/scss/_dropdown.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/Dropdown/scss/_dropdown.scss
@@ -34,6 +34,10 @@
 		.vscomp-wrapper {
 			font-family: inherit;
 		}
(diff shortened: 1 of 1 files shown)


gnbm:
- This is a customization that should be done at the consumer application level since is not according to our guidelines. 
- Because of that, I'll close this PR, but please keep exploring and contributing to our repo 🚀

alexandre97costa:
> * This is a customization that should be done at the consumer application level since is not according to our guidelines.


Where can I see these guidelines? From an outside view, it doesn't make sense to explicitly remove an accessibility feature. Besides, consistency across input types is lost when navigating with the keyboard: A text input gets focused and shows a primary border, but Tab-focusing on a dropdown does nothing, visually.

PR #855, opened 2023-11-08
ROU-4582:  SetBreakPoints - Remove wrong tablet validation when window height > width 

OS-giulianasilva:
This PR is to fix the SetDeviceBreakpoint assigning a desktop as a tablet.

### What was happening
-  When at desktop and if widowHeight <= tabletMax it will be set the tablet class, which is wrong.

### What was done
- Simplified the logic so it can work as expected:
    - The user sets the phoneWidth and tabletWidth breakpoints.
    - If both window dimensions are smaller than the phoneWidth, then it is a phone.
    - If both window dimensions are bigger than the phoneWidth and smaller than the tabletWidth, then it is a tablet.
    - If both window dimensions are bigger than the tabletWidth, then it is a desktop.
    - To achieve that, we just need to validate the biggest dimension:
        - If landscape, we check the width
        - If portrait, we check the height. 
- So these changes were done:
    - Created a new variable, windowSize, to get the largest size between windowWidth and windowHeight. If in landscape orientation, get the width, if in portrait orientation, get the height.
    - Used the windowSize to set the correct device .

### Screenshots
- Before:
wrongValidationSetDeviceBreakpoint
- After: 
validationSetDeviceBreakpoint

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/dist/OutSystemsUI.d.ts
+++ b/dist/OutSystemsUI.d.ts
@@ -57,6 +57,7 @@ declare namespace OSFramework.OSUI.Constants {
             True: string;
         };
     };
+    const AllowPropagationAttr = "[data-allow-event-propagation]";
     const Dot = ".";
     const Comma = ",";
     const EnableLogMessages = false;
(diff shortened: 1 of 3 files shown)


joselrio (approved):
💪

BenOsodrac (approved):

PR #861, opened 2023-11-24
ROU-4629: Fix an issue on datepicker to match values on InitialDate

bmarcelino-fe:
This PR is for fix an issue on datepicker to match values the platform input and Datepicker InitialDate

### What was happening
- The date that are being applied on platform input, doesn't match with datepicker's date and when the date format is different between server and datepicker.

### What was done
- Created a new method to check the InitialDate that needs to be applied on DatePicker
- Validate in platform input contains value and set as a InitialDate, if the DatePicker doens't have a date defined

### Test Steps

1. Open test page
2. Check if the date on input is the same of the platfomr hidden input
3. Open datepicker and select a new date
4. Check again the dates

### Checklist
-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/dist/OutSystemsUI.css
+++ b/dist/OutSystemsUI.css
@@ -507,6 +507,7 @@ a{
 a, a:visited{
   color:var(--color-primary);
   -webkit-transition:all 180ms linear;
+  -o-transition:all 180ms linear;
   transition:all 180ms linear;
 }
 a:hover, a:focus{
@@ -1105,6 +1106,7 @@ body,
   position:fixed;
   top:0;
   -webkit-transition:opacity 130ms ease-in;
+  -o-transition:opacity 130ms ease-in;
   transition:opacity 130ms ease-in;
   width:100vw;
   will-change:opacity;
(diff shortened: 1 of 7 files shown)


dist/OutSystemsUI.js
```diff
@@ -18346,10 +18350,29 @@ var Providers;
...
+                            if (clearPlatformInput) {
+                                OSFramework.OSUI.Helper.Dom.Attribute.Set(this.datePickerPlatformInputElem, OSFramework.OSUI.GlobalEnum.HTMLAttributes.Value, OSFramework.OSUI.Constants.EmptyString);
+                            }
+                            this.prepareConfigs();
+                        }
                         onDateSelectedEvent(selectedDates) {
                             let _selectedDate = '';
                             if (selectedDates.length > 0) {
-                                _selectedDate = this.provider.formatDate(selectedDates[0], this.flatpickrOpts.dateFormat);
+                                _selectedDate = this.provider.formatDate(selectedDates[0], this.provider.enableTime ? 'Y-m-d H:i:S' : 'Y-m-d');
```
bmarcelino-fe, line 18375: After spoke with @joselrio , we decided to keep this format as string and not on a Enum because this a specific use case on DatePicker SingleDate.

OS-giulianasilva (approved):

dist/OutSystemsUI.css
gnbm, line 1: @bmarcelino-fe As we already talked about in the past, please avoid committing the Dist files

PR #863, opened 2023-11-29
ROU-4654: Fix issue of Safari related to scrollend event

bmarcelino-fe:
This PR is for fix issue of Safari related to scrollend event

### What was happening
- On Safari or iOS, the event scrollend isn't supported.

### What was done
- Created a validation for safari and iOS to focus the element whitout scrollend event.

### Checklist
-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Helper/FocusFirstInvalidInput.ts
+++ b/src/scripts/OSFramework/OSUI/Helper/FocusFirstInvalidInput.ts
@@ -37,22 +37,29 @@ namespace OSFramework.OSUI.Helper {
 			isSmooth: boolean,
 			elementParentClass: string
 		): void {
-			// Set the scrollable element to add the ScrollEnd event
-			const activeScreenElement = Helper.Dom.ClassSelector(
-				document.body,
-				GlobalEnum.CssClassElements.ActiveScreen
-			);
-
-			// Set the temporary function for event of ScrollEnd, to focus on element after the scroll occur
-			const focusOnScrollEnd = () => {
-				element.focus();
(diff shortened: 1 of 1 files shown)


JoaoFerreira-FrontEnd (approved):

src/scripts/OSFramework/OSUI/Helper/FocusFirstInvalidInput.ts
```diff
@@ -37,22 +37,29 @@ namespace OSFramework.OSUI.Helper {
...
-				GlobalEnum.CssClassElements.ActiveScreen
-			);
+			const browser = OSFramework.OSUI.Helper.DeviceInfo.GetBrowser();
 
-			// Set the temporary function for event of ScrollEnd, to focus on element after the scroll occur
-			const focusOnScrollEnd = () => {
+			OutSystems.OSUI.Utils.ScrollToElement(element.id, isSmooth, 0, elementParentClass);
+
+			// Set the element focus directly because the event scrollend isn't supported by safari or iOS
+			if (browser === GlobalEnum.Browser.safari || OSFramework.OSUI.Helper.DeviceInfo.IsIos) {
```
joselrio, line 45: We should be adding here an extra validation in order to check if isSmooth was set as True, and in that case, raise a console warning about this limitation. We should do the same thing at the input description param at Module side.

bmarcelino-fe: Added the console.warn to explain the limitation and added it to the platform action.

joselrio (approved):

JoaoFerreira-FrontEnd (dismissed):

PR #870, opened 2023-12-06
ROU-4664: Prevent dom move when inside popup

BenOsodrac:
This issue happens only in ODC, as the new version of react-dom adds all events on the ReactContainer, instead of adding it on every widget. This causes an issue with widgets inside the DropdownServerSide, when placed inside the Popup. As the Popup is placed outside the reactContainer, together with the MoveElement we are doing on the Dropdown, it causes all widgets inside to loose the attached listeners. 

- Added method to check if an element is inside the Popup widget.
- Added exception to only use MoveElement on DropdownServerSide if its not inside Popup, otherwise add custom class
- Added different CSS with that class, to adjust behaviour for dropdown inside the Popup

This approach was followed, as no other tested solved the issue. Removing the MoveElement and changing the overall position to fixed, would cause issues inside other patterns, like the Sidebar  This was the most safe and reliable solution found.

### Checklist

-   [ ] tested locally
-   [ ] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Helper/Dom.ts
+++ b/src/scripts/OSFramework/OSUI/Helper/Dom.ts
@@ -446,11 +446,18 @@ namespace OSFramework.OSUI.Helper {
 		 * @memberof Dom
 		 */
 		public static IsInsidePopupWidget(element: HTMLElement): boolean {
-			const _popup = document.querySelector(Constants.Dot + GlobalEnum.CssClassElements.Popup);
+			const _popup = document.querySelectorAll(Constants.Dot + GlobalEnum.CssClassElements.Popup);
+			let _isInsidePopup = false;
 
-			if (_popup && element) {
-				return _popup.contains(element);
+			if (_popup.length > 0 && element) {
+				_popup.forEach((popup) => {
+					if (popup.contains(element)) {
+						_isInsidePopup = true;
+					}
+				});
 			}
+
+			return _isInsidePopup;
 		}
 
 		/**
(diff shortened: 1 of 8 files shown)


BenOsodrac:
Code-smells are unrelated to changed code.

OS-giulianasilva (approved):

OS-giulianasilva (approved):

src/scripts/OSFramework/OSUI/Helper/Dom.ts
```diff
@@ -437,6 +437,22 @@ namespace OSFramework.OSUI.Helper {
...
+		/**
+		 * Method to check if element is inside a Popup widget
+		 *
+		 * @static
+		 * @param {HTMLElement} element
+		 * @return {*}  {boolean}
+		 * @memberof Dom
+		 */
+		public static IsInsidePopupWidget(element: HTMLElement): boolean {
+			const _popup = document.querySelector(Constants.Dot + GlobalEnum.CssClassElements.Popup);
```
joselrio, line 449: Can't we've more than one popup? If so, we should look by the popup that's open/active, right?

src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/DropdownServerSide.ts
```diff
@@ -971,8 +972,20 @@ namespace OSFramework.OSUI.Patterns.Dropdown.ServerSide {
 			this._setUpEvents();
 			// Add CSS classes
 			this._setCssClasses();
-			// Ensure that the Move only happens after HTML elements has been set!
-			this._moveBallonElement();
+
+			// Check if the dropdown is placed inside a Popu Widget
```
joselrio, line 976: typo at "Popu"

joselrio (changes requested):

OS-giulianasilva (approved):

joselrio (approved):

PR #875, opened 2023-12-13
ROU-4645: Add a missing validation when range was not defined.

joselrio:
This PR is for fix an issue when an extensibility attribute is missing at RangeSlider by adding an extra validation before checking it's value.

### DIFF
--- a/src/scripts/Providers/OSUI/RangeSlider/NoUISlider/Utils.ts
+++ b/src/scripts/Providers/OSUI/RangeSlider/NoUISlider/Utils.ts
@@ -5,7 +5,7 @@ namespace Providers.OSUI.RangeSlider.NoUISlider.Utils {
 		const _noUiSliderConfigs = providerConfigs;
 
 		// Check if the range key is empty. If true, remove from the array
-		if (_noUiSliderConfigs.range !== undefined && _noUiSliderConfigs.range.length <= 0) {
+		if (_noUiSliderConfigs.range?.length <= 0) {
 			delete _noUiSliderConfigs.range;
 		} else {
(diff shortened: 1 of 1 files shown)


src/scripts/Providers/OSUI/RangeSlider/NoUISlider/Utils.ts
```diff
@@ -5,7 +5,7 @@ namespace Providers.OSUI.RangeSlider.NoUISlider.Utils {
 		const _noUiSliderConfigs = providerConfigs;
 
 		// Check if the range key is empty. If true, remove from the array
-		if (_noUiSliderConfigs.range.length <= 0) {
+		if (_noUiSliderConfigs.range !== undefined && _noUiSliderConfigs.range.length <= 0) {
```
BenOsodrac, line 8: You could simply do ` if (_noUiSliderConfigs.range?.length <= 0) {}`

BenOsodrac (approved):

src/scripts/Providers/OSUI/RangeSlider/NoUISlider/Utils.ts
joselrio: Indeed! at the end of compilation it will be the same... But more clean on our side yes! Thank you ;)

joselrio: Done!

BenOsodrac (dismissed):

OS-giulianasilva (approved):

BenOsodrac (approved):

PR #887, opened 2024-01-09
[ROU-3816 V2]: Fixed ScreenOnScroll and ToggleNativeBehavior API issues

joselrio:
**This PR will add fixes related with:**
> ToggleNativebehaviour API:
- Method was implemented in both DatePicker and DatePickerRange, which ends into an issue since the redraw was happening in both contexts when at DatePickerRange this will never happen.

> ScreenOnScroll event:
- Implemented logic to validate if ScreenOnScroll container has changed, if so, update the targetElement accordingly, this will happen at the screen transition when the container with scroll will be the same but not the same instance!

Also improved code readability on contexts mentioned above.

### DIFF
--- a/src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/ScreenOnScroll.ts
+++ b/src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/ScreenOnScroll.ts
@@ -10,17 +10,19 @@ namespace OSFramework.OSUI.Event.DOMEvents.Listeners {
 		let scrollableContainer: HTMLElement | Document = undefined;
 
 		// If native or pwa app when NOT android, scrollable container will be the .content inside .active-scren
-		if (
-			OSFramework.OSUI.Helper.DeviceInfo.IsAndroid === false &&
-			(OSFramework.OSUI.Helper.DeviceInfo.IsNative || OSFramework.OSUI.Helper.DeviceInfo.IsPwa)
-		) {
-			scrollableContainer = Helper.Dom.ClassSelector(
-				document,
-				`${GlobalEnum.CssClassElements.ActiveScreen} ${Constants.Dot}${GlobalEnum.CssClassElements.Content}`
-			);
-		} else {
-			// At non native or android apps, .active-screen is the container with scroll
-			scrollableContainer = Helper.Dom.ClassSelector(document, GlobalEnum.CssClassElements.ActiveScreen);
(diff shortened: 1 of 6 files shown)


src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/ScreenOnScroll.ts
```diff
@@ -9,18 +9,20 @@ namespace OSFramework.OSUI.Event.DOMEvents.Listeners {
...
-			(OSFramework.OSUI.Helper.DeviceInfo.IsNative || OSFramework.OSUI.Helper.DeviceInfo.IsPwa)
-		) {
-			scrollableContainer = Helper.Dom.ClassSelector(
-				document,
-				`${GlobalEnum.CssClassElements.ActiveScreen} ${Constants.Dot}${GlobalEnum.CssClassElements.Content}`
-			);
-		} else {
-			// At non native or android apps, .active-screen is the container with scroll
-			scrollableContainer = Helper.Dom.ClassSelector(document, GlobalEnum.CssClassElements.ActiveScreen);
+		// Check based on the OS since Helper.DeviceInfo.Is* are returnin false in all cases since body classes are not set when this will be executed.
```
BenOsodrac, line 12: returning*

src/scripts/OSFramework/OSUI/Event/DOMEvents/Listeners/ScreenOnScroll.ts
```diff
@@ -9,18 +9,20 @@ namespace OSFramework.OSUI.Event.DOMEvents.Listeners {
...
+			case GlobalEnum.MobileOS.Unknown:
+			case GlobalEnum.MobileOS.Windows:
+				scrollableContainer = Helper.Dom.ClassSelector(document, GlobalEnum.CssClassElements.ActiveScreen);
+				break;
+			case GlobalEnum.MobileOS.IOS:
+				scrollableContainer = Helper.Dom.ClassSelector(
+					document,
+					`${GlobalEnum.CssClassElements.ActiveScreen} ${Constants.Dot}${GlobalEnum.CssClassElements.Content}`
+				);
+				break;
```
BenOsodrac, line 25: This could probably be an If(). If it's Ios do something, otherwise do X

BenOsodrac (approved):

JoaoFerreira-FrontEnd (approved):

PR #903, opened 2024-02-01
ROU-4712: Dropdown Search closes unexpectedly when is used on an Android Tablet

bmarcelino-fe:
This PR is for fixing an issue on mobile devices that trigger the onresise event and closes unexpectedly the keyboard.

### What was happening
- When a DropdownSearch is opened on an tablet android device, the keyboard opens and closes unexpectedly because of the OnResize event that is triggered

### What was done
- Create a method IsMobileDevice that returns if is a mobile device
- Added an validation that uses the value of IsMobileDevice to add the event OnOrientationChange for mobile devices or the OnResize event to desktop applications

### Test Steps

1. Open sample page on an Android Device in landscape
2. Open a dropdown
3. Click on the input search to open keyboard (Expected: the keyboard will continue open)
4. Rotate device (Expected: The keyboard and dropdown will close)
5. Open a dropdown
6. Click on the input search to open keyboard (Expected: the keyboard will continue open)
7. Rotate device (Expected: The keyboard and dropdown will close)

### Checklist
-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/src/scripts/OSFramework/OSUI/Helper/Device.ts
+++ b/src/scripts/OSFramework/OSUI/Helper/Device.ts
@@ -461,7 +461,7 @@ namespace OSFramework.OSUI.Helper {
 		}
 
 		/**
-		 * Method to return if is a mobile device based on is operating system
+		 * Method to return if is running on a mobile device
 		 *
 		 * @readonly
 		 * @static
--- a/src/scripts/OSFramework/OSUI/Helper/Device.ts
+++ b/src/scripts/OSFramework/OSUI/Helper/Device.ts
@@ -461,6 +461,21 @@ namespace OSFramework.OSUI.Helper {
 		}
 
 		/**
+		 * Method to return if is a mobile device based on is operating system
+		 *
+		 * @readonly
+		 * @static
+		 * @type {boolean}
+		 * @memberof DeviceInfo
+		 */
+		public static get IsMobileDevice(): boolean {
+			const isMobileDevice =
+				DeviceInfo.GetOperatingSystem() === GlobalEnum.MobileOS.Android ||
+				DeviceInfo.GetOperatingSystem() === GlobalEnum.MobileOS.IOS;
+			return isMobileDevice;
+		}
+
+		/**
 		 * Gets the Notch Position.
 		 *
 		 * @private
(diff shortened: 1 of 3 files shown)


src/scripts/OSFramework/OSUI/Helper/Device.ts
```diff
@@ -460,6 +460,21 @@ namespace OSFramework.OSUI.Helper {
 			return DeviceInfo._isTouch;
 		}
 
+		/**
+		 * Method to return if is a mobile device based on is operating system
```
BenOsodrac, line 464: comment seems confusing

src/scripts/Providers/OSUI/Dropdown/VirtualSelect/AbstractVirtualSelect.ts
```diff
@@ -40,6 +42,13 @@ namespace Providers.OSUI.Dropdown.VirtualSelect {
 			}
 		}
 
+		// Method to close the Dropdown if it's opened
+		private _closeOpened(): void {
```
BenOsodrac, line 46: Sorry to be picky about names, but this one could be better 😂 

Why not simply _close? 

Or do we need this method, instead of using the existing one, with this validation?

BenOsodrac (changes requested):

src/scripts/Providers/OSUI/Dropdown/VirtualSelect/AbstractVirtualSelect.ts
bmarcelino-fe: @bmarcelino-fe this method is when the events `onorientationchange` and `onresize` is triggered. Can't be simply close because this method only do it if it's open. The existing method close is used on client action that is public and needs to pass parameters whith an async. After talking with @joselrio we decided to create a this method to be used only on the events.

src/scripts/OSFramework/OSUI/Helper/Device.ts
bmarcelino-fe: @BenOsodrac I can improve the comment, however this method is to check if we are in a mobile device. The validation are based on device operating system, that will returns `ios` or `android` in nowadays.

BenOsodrac (approved):

joselrio (approved):

PR #909, opened 2024-02-14
Applying prettier configs to project

rugoncalves:
This PR is for applying the prettier configs to the project, so that the PRs are not polluted with the new configs.

### What was happening
* The prettier configs were changed
* This can cause that when a file is changed, the formatting will be changed
* The respective PR of a given task will bring many formatting changed, causing the actual changes to be hidden

### What was done
* Ran the command: `npm run prettier`
* All changes caused by it, being now commit

### Test Steps
1. No actual code change
2. No need to test code

### Checklist
* [ ] tested locally
* [ ] documented the code
* [x] clean all warnings and errors of eslint

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -28,5 +28,9 @@
 	},
 	"[jsonc]": {
 		"editor.defaultFormatter": "vscode.json-language-features"
+	},
+	"sonarlint.connectedMode.project": {
+		"connectionId": "outsystems",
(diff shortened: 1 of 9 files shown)


gnbm (approved):
LGTM 🚀 
@rugoncalves probably we'll need to check the other repos, right?

BenOsodrac (approved):

PR #919, opened 2024-03-08
Update version at missing file.

joselrio:
This PR will update version into a missing bumped version file.
This file will be responsible to set the version at the compiled files and also at the generated themes for both versions.


OS-giulianasilva (approved):

gnbm (approved):
LGTM

PR #925, opened 2024-03-25
ROU-4807: fix transform property

BenOsodrac:
This PR is for fixing the scss compilation for the scaleup animation.

The issue with the screen-transition scss compilation (transition with multiple properties resulta in 3 transition properties being compiled). From my research and attempts, this seems an issue from scss and was not able to find any workaround. That being said, in runtime it will still work as expected.

### Checklist

-   [x] tested locally
-   [x] documented the code


gnbm (approved):
LGTM 👍🏻

OS-giulianasilva (approved):

PR #926, opened 2024-03-26
ROU-4817: Fix pickers inside Sidebar not following scroll

OS-giulianasilva:
This PR is for fixing the pickers inside the Sidebar not following scroll.

### What was happening
- When inside a Sidebar, the calendar position was not updated when scrolling.

### What was done
- Added a scroll event listener to the Sidebar element that updates the calendar position on scroll.

### Checklist
-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/Providers/OSUI/SharedProviderResources/Flatpickr/UpdatePositionOnScroll.ts
+++ b/src/scripts/Providers/OSUI/SharedProviderResources/Flatpickr/UpdatePositionOnScroll.ts
@@ -4,7 +4,9 @@ namespace Providers.OSUI.SharedProviderResources.Flatpickr {
 		// Indicates wether the picker is a time picker or other type of picker (Month, Date, DateTime).
 		private _isTimePicker: boolean;
 		// Event OnScreenScroll
-		private _onScreenScrollEvent: OSFramework.OSUI.GlobalCallbacks.Generic;
+		private _onScrollEvent: OSFramework.OSUI.GlobalCallbacks.Generic;
+		// Store the parent Sidebar instance if any
+		private _parentSidebar: HTMLElement;
 		// Store the picker instance
 		private _picker:
 			| Datepicker.Flatpickr.IFlatpickr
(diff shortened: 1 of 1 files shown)


src/scripts/Providers/OSUI/SharedProviderResources/Flatpickr/UpdatePositionOnScroll.ts
```diff
@@ -4,7 +4,9 @@ namespace Providers.OSUI.SharedProviderResources.Flatpickr {
 		// Indicates wether the picker is a time picker or other type of picker (Month, Date, DateTime).
 		private _isTimePicker: boolean;
 		// Event OnScreenScroll
-		private _onScreenScrollEvent: OSFramework.OSUI.GlobalCallbacks.Generic;
+		private _onScrollEvent: OSFramework.OSUI.GlobalCallbacks.Generic;
+		// Store the parent Sidebar instance if any
+		private _parentSidebar: HTMLElement;
```
gnbm, line 9: @OS-giulianasilva couldn't we have this more abstract and not directly coupled with the Sidebar? I was thinking that we can have other elements/components that will fit in the same pocket

BenOsodrac (approved):

src/scripts/Providers/OSUI/SharedProviderResources/Flatpickr/UpdatePositionOnScroll.ts
OS-giulianasilva: Not sure how to do it, to be honest. I need to find this pattern and add the scroll event listener to it, but first I need to know which pattern it is so I can find it using the `closest` method. 
If the pattern were always the picker's direct parent, it would be easier, but this is not always true.

gnbm: Let's move forward @OS-giulianasilva Is something we can try to iterate later on ;)

gnbm (approved):

PR #943, opened 2024-05-13
ROU-4770: Fix Dropdown Server Side issues after integrating the OSUI Balloon

OS-giulianasilva:
This PR is for fixing Dropdown Server Side issues after integrating the OSUI Balloon

### What was happening
- Inside new Tabs and on some Tab contents and not in RTL the right border is not displayed: 
image
- Inside new Tabs when the dropdown is open and we click on another Tab without closing the dropdown balloon, we temporarily see the balloon flickering on the left side of the screen: 
IssueDropdownSerSideInsideNewTabs

### What was done
- In the Dropdown Server Side, subtracted the size of its border from the width so it is displayed correctly inside the Tabs.
- In the Tabs, added "opacity:0" to the not active tabs content items.

### Checklist

-   [x] tested locally
-   [x] documented the code


JoaoFerreira-FrontEnd (approved):

gnbm (approved):
LGTM

PR #948, opened 2024-05-23
ROU-4825: Prevent multiple redraws.

joselrio:
This PR is will improve redraw process in order to prevent multiple redraws.


BenOsodrac (dismissed):

src/scripts/OSFramework/OSUI/Pattern/AbstractProviderPattern.ts
```diff
@@ -117,6 +121,23 @@ namespace OSFramework.OSUI.Patterns {
...
+		 */
+		protected redraw(): void {
+			/* If a redraw has been trigger before the previous occured, 
+			cancel the previous one and restart a new redraw process, 
+			this way multiple redraws will be prevented.  */
+			this._cancelRedraw();
+
+			this._refreshTimeOutId = window.setTimeout(() => {
+				this._handleRedraw();
+			}, 0);
```
rugoncalves, line 138: I would prefer us to use the helper `AsyncInvocation`. For that we'd need to change the helper action to return the id of the timer. But It think it would be cleaner or more obvious.

src/scripts/OSFramework/OSUI/Pattern/AbstractProviderPattern.ts
```diff
@@ -35,6 +37,13 @@ namespace OSFramework.OSUI.Patterns {
 			this.isProviderBased = true;
 		}
 
+		// Method that will check if there is already a redraw process runing in order to cancel it
+		private _cancelRedraw(): void {
+			if (this._refreshTimeOutId !== undefined) {
```
rugoncalves, line 42: we're not resetting the attribute `this._refreshTimeOutId` to `undefined`...

rugoncalves (changes requested):

src/scripts/OSFramework/OSUI/Pattern/AbstractProviderPattern.ts
joselrio: well noted! will do it! ;)

src/scripts/OSFramework/OSUI/Pattern/AbstractProviderPattern.ts
joselrio: 🤦 Thank you!

src/scripts/OSFramework/OSUI/Pattern/AbstractProviderPattern.ts
joselrio: Done!

src/scripts/OSFramework/OSUI/Pattern/AbstractProviderPattern.ts
joselrio: Done!

BenOsodrac (approved):

rugoncalves (approved):

PR #953, opened 2024-06-06
ROU10687: DropdownServerSide style issues

joselrio:
This PR will:
- Fix a style issue that was preventing DropdownServerSide to be set with no-valid (error) style;
> Before:
Screenshot 2024-06-06 at 17 34 49

> After:
Screenshot 2024-06-06 at 17 35 22

- Enhance A11Y outline style to the DropdownServerSide options ballon when it's focused;
> Before:
Screenshot 2024-06-06 at 17 35 02

> After:
Screenshot 2024-06-06 at 17 35 34

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/scss/_dropdown-serverside.scss
+++ b/src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/scss/_dropdown-serverside.scss
@@ -227,7 +227,8 @@ $balloonMobileTopMargin: 5vh;
 	// When it's not valid
 	&--not-valid {
 		.osui-dropdown-serverside__selected-values-wrapper {
-			border-color: var(--color-error);
+			/* !important should be here in order to grant more specific selectors be overriding border color at this context. */
+			border-color: var(--color-error) !important;
 		}
 
 		& + .osui-dropdown-serverside-error-message {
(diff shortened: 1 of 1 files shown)


src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/scss/_dropdown-serverside.scss
```diff
@@ -227,7 +228,8 @@ $balloonMobileTopMargin: 5vh;
 	// When it's not valid
 	&--not-valid {
 		.osui-dropdown-serverside__selected-values-wrapper {
-			border-color: var(--color-error);
+			/* !important should be here in order to grant more specific selectors be overriding border color at this context. */
+			border-color: var(--color-error) !important;
```
BenOsodrac, line 232: Personally, I would avoid it and just increase specificity if needed, but will approve the PR

BenOsodrac (approved):

src/scripts/OSFramework/OSUI/Pattern/Dropdown/ServerSide/scss/_dropdown-serverside.scss
joselrio: I understand your point @BenOsodrac, however this is an issue when **has-accessible-features selector** is enabled since the new border color for the A11Y will be more specific which ends on not having the expected border color (not-valid). 
On that case we need to choose between redefine this color when A11Y is enabled, or this one...

JoaoFerreira-FrontEnd (approved):

PR #963, opened 2024-07-03
ROU-10851: fix AnimatedLabel on init value

BenOsodrac:
This PR is for fixing the issue that prevents the AnimatedLabel to correctly appear as active on the screen load, when there's already a default value.

This was happening as the build method was async, while the render method responsible for the active state was running first than the build. As such, the render method was also made async.

### Checklist

-   [x] tested locally
-   [ ] documented the code


BenOsodrac:
Pipeline failure fixed will be fixed by: https://github.com/OutSystems/outsystems-ui-tests/pull/454

rugoncalves (approved):

joselrio (approved):

PR #968, opened 2024-07-15
ROU-10919: Fixed how to get URL file extension for video component

gnbm:
This PR is for fixing an issue that caused the Video component not to play as expected.

### What was happening

- This occurred when using a URL from a resource within the app since the URL gets appended with a session hash, which prevented the video extension from being properly calculated.

image

### What was done

- Created a method to properly calculate the file extension for a given URL despite the fact of having a session hash or not.

### Test Steps

1. Create a sample screen with an instance of video using a URL from a resource (e.g. `Resources.osuivideo_mp4.URL`) 
2. Check that the video works
3. Added an instance of video using a URL from an external source (e.g. `http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/Sintel.mp4`)
4. Check that the video works

### Screenshots

image

### Checklist

-   [ ] tested locally
-   [ ] documented the code

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+# Editor configuration, see https://editorconfig.org
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 2
--- a/src/scripts/OSFramework/OSUI/Helper/URL.ts
+++ b/src/scripts/OSFramework/OSUI/Helper/URL.ts
@@ -3,6 +3,20 @@ namespace OSFramework.OSUI.Helper {
 	// eslint-disable-next-line @typescript-eslint/naming-convention
 	export abstract class URL {
 		/**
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,6 @@
+# Editor configuration, see https://editorconfig.org
+[*]
+charset = utf-8
```
BenOsodrac, line 3: @gnbm was this intented?

BenOsodrac (approved):

JoaoFerreira-FrontEnd (approved):

.editorconfig
gnbm: It was. To make sure the editor configs work equally between Mac and Windows..hopefully
Did you find any impact on this?

BenOsodrac: No, just curious 👍

PR #969, opened 2024-07-16
ROU-10923: DatePicker TodayButton didn't have tabindex=-1 when closed

gnbm:
### What was happening

- Issue where the "Today's" button in the DatePicker/DatePickerRange components did not have a `tabindex = -1` when the DatePicker was closed.
- This occurred when the ShowTodayButton option was set to True and the DatePicker/DatePickerRange instance was closed.

### What was done

- Changed the OnClose handler to set properly `tabindex = -1` when closing the DatePicker/DatePickerRange.
- Added a new OnOpen handler to set properly `tabindex = 0` when opening the DatePicker/DatePickerRange.
- Changed `addTodayBtn()` to set `tabindex = -1` when opened and `tabindex = 0` when closed.

### Test Steps

1. Go to the DatePicker sample screen
2. Enable ShowTodayButton
3. Clicked on Apply 
4. Open the DatePicker 
5. Check the today button works and has `tabindex = 0`
6. Close the Date Picker
7.  Check the Today button has `tabindex = -1`
8. Repeat the same test for DatePickerRange

### Checklist

-   [X] tested locally
-   [X] documented the code
-   [ ] requires changes in OutSystems 
-   [ ] requires new sample page in OutSystems

### DIFF
--- a/src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
+++ b/src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
@@ -23,6 +23,8 @@ namespace Providers.OSUI.Datepicker.Flatpickr {
 		protected onSelectedCallbackEvent: OSFramework.OSUI.Patterns.DatePicker.Callbacks.OSOnChangeEvent;
 		// Property to store a custom callback called on OnClose Flatpickr event
 		public onCloseCustomCallback = undefined;
+		// Property to store a custom callback called on OnOpen Flatpickr event
+		public onOpenCustomCallback = undefined;
 
 		constructor(uniqueId: string, configs: C) {
 			super(uniqueId, configs);
@@ -32,6 +34,8 @@ namespace Providers.OSUI.Datepicker.Flatpickr {
 
 			// Set the default library Event handler that will be used to set on the provider configs
 			this.configs.OnClose = this.onDatePickerClose.bind(this);
+
+			this.configs.OnOpen = this.onDatePickerOpen.bind(this);
 		}
 
 		// Method used to set the needed HTML attributes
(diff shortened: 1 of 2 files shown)


BenOsodrac (approved):

BenOsodrac (approved):

src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
```diff
@@ -249,6 +266,28 @@ namespace Providers.OSUI.Datepicker.Flatpickr {
...
+				OSFramework.OSUI.Helper.A11Y.TabIndexTrue(
+					this._todayButtonElem.querySelector(OSFramework.OSUI.GlobalEnum.HTMLElement.Link)
+				);
+			}
+			// Call the custom callback provided for users to add custom code when a datepicker opens
+			if (
+				this.onOpenCustomCallback !== undefined &&
+				typeof this.onOpenCustomCallback === OSFramework.OSUI.Constants.JavaScriptTypes.Function
+			) {
+				this.onOpenCustomCallback(this.provider);
```
rugoncalves, line 287: This a new feature, correct?
If so, should it be implemented in the context of this task?
Also, shouldn't it be asynchronously invoked?

gnbm: @rugoncalves Adding this new event was the only way I found to change the `tabindex` while opening the date pickers (after the first time) and followed the same approach used by OnClose to call the provider's event.
About being asynchronous why would we need that?

gnbm: @BenOsodrac any context on this since I followed the same approach as for `OnClose`? 
I assume that it was implemented thinking only about an internal usage of the events and so it's synchronous, but have no context.

BenOsodrac: I think its ok yes, close and open are just internal events on JS, so no need to be async.

rugoncalves (approved):

BenOsodrac (approved):

PR #984, opened 2024-08-29
ROU-11005: DatePicker, DatePickerRange, MonthPicker and TimePicker

joselrio:
### What was happening
- It was possible to focus the Input during tabNavigation even when disabled.

### What was done
- Created a new OnRender API that will check if the input disabled attribute has been updated, if so, tabIndex and aria-hidden attributes of the visible input will be updated accordingly.

### DIFF
--- a/src/scripts/OutSystems/OSUI/Patterns/TimePickerAPI.ts
+++ b/src/scripts/OutSystems/OSUI/Patterns/TimePickerAPI.ts
@@ -182,7 +182,7 @@ namespace OutSystems.OSUI.Patterns.TimePickerAPI {
 		const result = OutSystems.OSUI.Utils.CreateApiResponse({
 			errorCode: ErrorCodes.TimePicker.FailOnRender,
 			callback: () => {
-				const _timePickerItem = GetTimePickerItemById(timePickerId);
+				const _timePickerItem = this.GetTimePickerItemById(timePickerId);
 
 				_timePickerItem.onRender();
 			},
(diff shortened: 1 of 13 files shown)


BenOsodrac (approved):

src/scripts/OutSystems/OSUI/Patterns/TimePickerAPI.ts
```diff
@@ -171,6 +171,26 @@ namespace OutSystems.OSUI.Patterns.TimePickerAPI {
...
+	 *
+	 * @export
+	 * @param {string} timePickerId
+	 * @return {*}  {string}
+	 */
+	export function OnRender(timePickerId: string): string {
+		const result = OutSystems.OSUI.Utils.CreateApiResponse({
+			errorCode: ErrorCodes.TimePicker.FailOnRender,
+			callback: () => {
+				const _timePickerItem = GetTimePickerItemById(timePickerId);
```
OS-giulianasilva, line 185: The other APIs are using `this.GetTimePickerItemById`. Shouldn't this one be like this too?

OS-giulianasilva (approved):

src/scripts/OutSystems/OSUI/Patterns/TimePickerAPI.ts
joselrio: nice catch 💪

BenOsodrac (approved):

PR #985, opened 2024-09-03
ROU-11113: DatePicker and DatePickerRange, Fix validation message position.

joselrio:
This PR will fix the validation message position once inside an Input With Icon pattern.

### DIFF
--- a/src/scss/04-patterns/03-interaction/_input-with-icon.scss
+++ b/src/scss/04-patterns/03-interaction/_input-with-icon.scss
@@ -69,8 +69,6 @@
 		&.validation-message {
 			display: block;
 			position: absolute;
-			// the !important is used to override a more specific style containing ThemeGrid_Width context which is not needed here
-			left: 0 !important;
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


src/scss/04-patterns/03-interaction/_input-with-icon.scss
```diff
@@ -69,6 +69,8 @@
 		&.validation-message {
 			display: block;
 			position: absolute;
+			// the !important is used to override a more specific style containing ThemeGrid_Width context which is not needed here
+			left: 0 !important;
```
BenOsodrac, line 73: Then let's make this rule more specific, is still better than !important

BenOsodrac (changes requested):

src/scss/04-patterns/03-interaction/_input-with-icon.scss
joselrio: Mixed feelings about it... but's done... ;)

BenOsodrac (approved):

PR #998, opened 2024-10-15
ROU-11295: Fix Tabs issue where getting focused while entering the screen

gnbm:
### What was happening

- When entering a screen, the Tabs component received focus unexpectedly when navigating to a screen. 
This occurred when having content positioned above the Tabs causing the page to scroll and bringing the Tabs into focus.

https://github.com/user-attachments/assets/9b998c5b-73d8-43a9-a0cb-22a1227d1e4a

### What was done

- A new validation was added to make sure the logic for focusing Tabs headers only happens once Tabs are built.
- Fixed a Snyk finding where we should be using `Number.isNaN(...)` instead of `isNaN(...)`.

### Test Steps

1. Go to sample screen 1 by clicking on a link from the home screen
2. Check that the screen remains at the top
3. Scroll down to the Tabs
4. Change to another Tab by clicking on the header 
5. Check that the tab that was opened got focused as expected
6. Clicking on a button calling the client action `SetActiveTab`
7. Check that the target tab that was opened got focused as expected

### Screenshots

Workaround

### Checklist

-   [X] tested locally
-   [X] documented the code

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,6 @@
 		"typedoc-plugin-merge-modules": "^4.0.1",
 		"typedoc-umlclass": "^0.7.0",
 		"typescript": "^4.5.0",
-		"virtual-select-plugin": "^1.0.45"
+		"virtual-select-plugin": "^1.0.46"
 	}
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -67,6 +67,6 @@
 		"typedoc-plugin-merge-modules": "^4.0.1",
 		"typedoc-umlclass": "^0.7.0",
 		"typescript": "^4.5.0",
-		"virtual-select-plugin": "^1.0.45"
+		"virtual-select-plugin": "^1.0.46"
```
BenOsodrac, line 70: Did you intend to upgrade the virtual select library on this PR?

BenOsodrac (approved):

joselrio (approved):

package.json
joselrio: 👆

PR #1011, opened 2025-01-23
ROU-11536: Rating - Fixed A11Y issue at the placeholder elements.

joselrio:
This PR will fix an A11Y issue at Rating pattern.

### What was happening

- Rating placeholders was being cloned and their Ids was being kept the same for each rating element, which causes a bad semantic structure due the usage of duplicated Ids.

### What was done

- Updated the way elements are being cloned and make sure Id is updated for each element.

### DIFF
--- a/src/scripts/OSFramework/OSUI/Pattern/Rating/Rating.ts
+++ b/src/scripts/OSFramework/OSUI/Pattern/Rating/Rating.ts
@@ -470,9 +470,9 @@ namespace OSFramework.OSUI.Patterns.Rating {
 			let placeholdersOuterHTML = '';
 
 			// Iterate over the cloned placeholders to redefine the id attribute
-			for (const placholder of this._clonedPlaceholders) {
+			for (const placeholder of this._clonedPlaceholders) {
 				// Get a clone of the placeholder to avoid changing the original (HTML Element)
-				const clonedPlaceholder = placholder.cloneNode(true) as HTMLElement;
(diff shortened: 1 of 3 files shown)


src/scripts/OSFramework/OSUI/Pattern/Rating/Rating.ts
```diff
@@ -456,6 +457,32 @@ namespace OSFramework.OSUI.Patterns.Rating {
...
+		 * @memberof Rating
+		 */
+		private _updateClonePlaceholdersAttrs(ratingIndex: number): string {
+			// Store the ourterHTML of the placeholders
+			let placeholdersOuterHTML = '';
+
+			// Iterate over the cloned placeholders to redefine the id attribute
+			for (const placholder of this._clonedPlaceholders) {
+				// Get a clone of the placeholder to avoid changing the original (HTML Element)
+				const clonedPlaceholder = placholder.cloneNode(true) as HTMLElement;
```
JoaoFerreira-FrontEnd, line 475: fix placeholder typo

JoaoFerreira-FrontEnd (changes requested):
Please check the typo

src/scripts/OSFramework/OSUI/Pattern/Rating/Rating.ts
joselrio: Done!

JoaoFerreira-FrontEnd (approved):
LGTM

PR #1022, opened 2025-03-03
ROU-11689: DatePicker - Fixed a DatePiker position issue

joselrio:
This PR will fix a position issue at the DatePicker.

### What was happening
- Once inside a Popup, calendar was not being properly set with the classes and position issues was happening...
Screenshot 2025-03-03 at 11 34 15

### What was done
- Removed a duplicated validation that was preventing DatePicker be set with the correct selectors.

### DIFF
--- a/src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
+++ b/src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
@@ -206,27 +206,19 @@ namespace Providers.OSUI.Datepicker.Flatpickr {
 
 			// Since native behaviour could be enabled, check if the calendar container exist!
 			if (this.provider.calendarContainer !== undefined) {
-				if (
-					this.configs.DisableMobile === true ||
-					OSFramework.OSUI.Helper.DeviceInfo.IsDesktop ||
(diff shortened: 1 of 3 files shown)


src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
```diff
@@ -206,27 +206,19 @@ namespace Providers.OSUI.Datepicker.Flatpickr {
...
-					OSFramework.OSUI.Helper.DeviceInfo.IsDesktop ||
-					this.configs.CalendarMode === OSFramework.OSUI.Patterns.DatePicker.Enum.Mode.Range
-				) {
-					// Add TodayBtn
-					if (this.configs.ShowTodayButton) {
-						this.addTodayBtn();
-					}
+				// Add TodayBtn
+				if (this.configs.ShowTodayButton) {
+					this.addTodayBtn();
```
BenOsodrac, line 211: Shouldn't we prevent the today button logic to run when the DatePicker native is being used? (DisableMobile === true)

joselrio: Yes, but we're already... the `this.provider.calendarContainer` only exist if not native, that's why other validations can be removed ;)

BenOsodrac (approved):

PR #1034, opened 2025-04-17
ROU-11827: SubMenu, Date and Time Pickers A11Y improvements.

joselrio:
This PR will add several A11Y improvements to:
- DatePicker
- TimePicker
- SubMenu

### What was happening

- **DatePicker**: Today button was being set as "disabled" for keyboard navigation due tabindex being set as -1 when picker was set to be used inline.

- **TimePicker**: Input Label issue - Even when developer added and set the Label to the input, Label was not being proper connected with it, which ends on A11Y reported issue mentioning a label should be set to each input.

- **SubMenu**: Aria attributes (role values) was being used in a wrong way.

### What was done

- Fixed the issues mentioned above.

### Checklist

-   [x] tested locally
-   [x] documented the code


src/scripts/Providers/OSUI/Timepicker/Flatpickr/FlatpickrTime.ts
```diff
@@ -58,6 +58,35 @@ namespace Providers.OSUI.TimePicker.Flatpickr {
...
+			 */
+			const inputLabelElem = OSFramework.OSUI.Helper.Dom.TagSelectorAll(
+				document.body,
+				OSFramework.OSUI.GlobalEnum.HTMLElement.Label
+			).filter(
+				(_lbl) =>
+					_lbl
+						.getAttribute(OSFramework.OSUI.GlobalEnum.HTMLAttributes.For)
+						.indexOf(this.timePickerPlatformInputElem.id) > -1
+			);
```
OS-giulianasilva, line 75: Can't you use the `querySelector` with a more specific selector instead of getting all the labels and filtering them?

OS-giulianasilva (changes requested):

src/scripts/Providers/OSUI/Datepicker/Flatpickr/AbstractFlatpickr.ts
```diff
@@ -505,7 +505,7 @@ namespace Providers.OSUI.Datepicker.Flatpickr {
 		public close(): void {
 			if (this.provider.isOpen) {
 				this.provider.close();
-				if (this.configs.ShowTodayButton) {
+				if (this.configs.ShowTodayButton && this.provider.config.inline === false) {
```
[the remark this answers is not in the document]
joselrio: No we can't, only because Today button is something we're adding through OutSystemsUI and not available at the provider. That's why this is being done under this context.

src/scripts/Providers/OSUI/Timepicker/Flatpickr/FlatpickrTime.ts
joselrio: No, Labels are being added and set by the developer. We do not have another way to check for the "linked" label.

OS-giulianasilva: Something like this wouldn't work? 
```typescript
document.querySelector(`${OSFramework.OSUI.GlobalEnum.HTMLElement.Label}[${OSFramework.OSUI.GlobalEnum.HTMLAttributes.For}=${this.timePickerPlatformInputElem.id}]`)
```

joselrio: Performance wise, this is a best approach when compared with the **querySelector**.

OS-giulianasilva (approved):

OS-giulianasilva (approved):

PR #1053, opened 2025-08-26
ROU-12149: [Menu] - Tabindex incorrectly added to nav tag in common/menu when window is maximized

JoaoFerreira-FrontEnd:
This PR is for fixing the tabIndex attribute after the screen is resized. 

### What was happening

- When the menu was on a screen that suffered a resize, there were situations where the tabIndex was not updated according the right deviceType;
- The code was looking into the deviceType class before it was updated by the client runtime code (300ms).  

### What was done
- confirm the internal appProp object is updated on showMenu;
- run onResizeCallback only after 300ms , to guarantee layout is updated;

### Test Steps

1. Open Sample
2. Resize it for mobile
3. Use TAB navigation to open the menu
4. Maximize the window by clicking on the window header or the maximize icon
5. Check if the <nav> Element (class: app-menu-content) loses the tabindex and aria-expanded attributes 

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/OutSystems/OSUI/Utils/Menu.ts
+++ b/src/scripts/OutSystems/OSUI/Utils/Menu.ts
@@ -2,6 +2,8 @@
 namespace OutSystems.OSUI.Utils.Menu {
 	// OrientationChange callback to be stored and removed on Destroy
 	let _onOrientationChangeCallback: OSFramework.OSUI.GlobalCallbacks.Generic;
+	// Store the OnResize settimeout, to be cleared when needed
+	let _onResizeTimeout: number;
 
 	// App Properties
 	const _appProp = {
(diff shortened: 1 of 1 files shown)


src/scripts/OutSystems/OSUI/Utils/Menu.ts
```diff
@@ -123,31 +125,34 @@ namespace OutSystems.OSUI.Utils.Menu {
...
-		// Remove the menu event listeners since device type changed
-		_removeMenuEventListeners();
+			// Remove the menu event listeners since device type changed
+			_removeMenuEventListeners();
 
-		// ReAdd the menu event listeners
-		_addMenuEventListeners();
+			// ReAdd the menu event listeners
+			_addMenuEventListeners();
+		}, 300);
```
joselrio, line 155: @JoaoFerreira-FrontEnd, I would say we should left a comment here about why these 300ms are needed here.

joselrio (approved):

src/scripts/OutSystems/OSUI/Utils/Menu.ts
JoaoFerreira-FrontEnd: added;

PR #1055, opened 2025-10-03
ROU-12253: Update Virtual Select to v1.1.2

gnbm:
This PR is to update the VirtualSelect library to the latest pre-release **v1.1.2**

### What was done

- Updated Virtual Select provider to the latest pre-release **v1.1.2**
    - _Note:_  couldn’t create an npm since the GH failed for permission, and [this](https://github.com/OutSystems/outsystems-ui/blob/b207c90e3391e7030e3e252a8b53ee34f53a609e/package.json#L70 ) will be outdated   

### Checklist

-   [X] tested locally
-   [X] documented the code
-   [X] requires changes in OutSystems (if so, provide a module with changes)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
 		"@typescript-eslint/eslint-plugin": "^5.62.0",
 		"@typescript-eslint/parser": "^5.62.0",
 		"branch-name": "^1.0.0",
-		"browser-sync": "^3.0.3",
+		"browser-sync": "^3.0.4",
 		"eslint": "^8.57.1",
 		"eslint-config-prettier": "^9.1.0",
 		"eslint-plugin-prettier": "^5.2.5",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
 		"@typescript-eslint/eslint-plugin": "^5.62.0",
 		"@typescript-eslint/parser": "^5.62.0",
 		"branch-name": "^1.0.0",
-		"browser-sync": "^3.0.3",
+		"browser-sync": "^3.0.4",
 		"eslint": "^8.57.1",
```
BenOsodrac, line 40: @gnbm missing the virtual-select udpate on the dev dependency right?

BenOsodrac (changes requested):

BenOsodrac (approved):

package.json
gnbm: @BenOsodrac, I left a comment in the PR and in the 'What I Did' section. We basically aren’t able to do npm releases in Virtual Select since the token expired, and the repo owner is unresponsive. 
Because of this, there’s no version in npm to be updated 😕

PR #1062, opened 2025-11-17
ROU-12315: Implement the changes needed to support edge-to-edge

rugoncalves:
From android version 16+ we are unable to opt out of edge to edge enforcement. This means that all applications targeting android SDK 36+ will take up the entire screen. This has a couple of effects, namely the removal of the status bar and the safe zone around the gesture/navigation bar. Instead, they are transparent, and you are able to display content inside of them. In terms of OutSystems Apps, we somewhat support this by injecting a CSS variable for the status bar that allows us to offset the content of the webview so it isn’t occluded by the icons of the status bar. This, however, does not exist for the navigation bar. This means that all existing OutSystems applications do not display correctly when running on an Android 16 device.

This PR is for updating the CSS variables being used for setting up the notch value in Android devices.

### What was happening
- Starting from Android 16, the applications would appear broken;
- MABS12 will introduce new CSS variables to be used;
- Retro-compatibility is required so old variables cannot be replaced;

### What was done
- Removed usage of attribute `data-status-bar-height`, as it will no longer exist, in newer versions;
- Replace usage of css variable `--status-bar-height` by `max(var(--status-bar-height), var(--safe-area-inset-top, 0px))`, which is retro-compatible;

### Test Steps
1. Open app in Android
2. Check that the top bar is not hidden
3. Toggle the status bar visibility with the new API
4. Confirm that the visibility state

### Screenshots
Status-bar-android-16

### Checklist
-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scss/01-foundations/_root.scss
+++ b/src/scss/01-foundations/_root.scss
@@ -83,10 +83,17 @@
 	--footer-height: #{$footer-height};
 
 	/*! iOS Safe Areas (Notch Support) */
-	--os-safe-area-top: env(safe-area-inset-top);
-	--os-safe-area-right: env(safe-area-inset-right);
-	--os-safe-area-bottom: env(safe-area-inset-bottom);
-	--os-safe-area-left: env(safe-area-inset-left);
+	/*! Before MABS12, the insets were obtained via env(safe-area-inset-*)
+		for iOS, as for Android the values were reported incorrectly.
+	    From MABS12 onwards, the native shell is injecting the CSS variables
+		"--safe-area-inset-*".
+		To ensure retro-compatibility, we use max() to get the highest value
+		between env() and the CSS variable.
+	*/
+	--os-safe-area-top: #{safe-area(top)};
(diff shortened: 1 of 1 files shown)


BenOsodrac (changes requested):
@rugoncalves Maybe its worth to create an internal scss variable that holds the `max(var(--status-bar-height), var(--safe-area-inset-top, 0px)),` that is repeated a lot and probably the root for the sonarcloud issue.

rugoncalves:
> @rugoncalves Maybe its worth to create an internal scss variable that holds the `max(var(--status-bar-height), var(--safe-area-inset-top, 0px)),` that is repeated a lot and probably the root for the sonarcloud issue.

That was the initial approach, unfortunately it doesn't work, as it is a known CSS limitation. The max() function in CSS cannot dynamically re-evaluate when custom properties change after the initial computation. CSS functions like max(), min(), clamp(), and calc() are computed at specified-value time, not live.

Unless you see some solution to this, we're stuck in repeating the expression everywhere, so that it may be recalculated.

PR #1149, opened 2026-06-09
ROU-12834: Add correct dropdown error placement

os-davidlourenco:
This PR aims to fix the misplacement of the arrow inside the dropdown on ODC.

### What was happening

- In ODC, when the dropdown is used inside a form, and the width is set to a col value, the dropdown arrow becomes misplaced.

### What was done

- An anchor name was defined on the dropdown select
- The anchor name was then used to place the arrow in the correct place

### Test Steps

1. Open the sample application
2. Validate that the dropdown has the arrow on the right place

### Checklist

-   [ ] tested locally
-   [ ] documented the code

### DIFF
--- a/src/scss/01-foundations/_icon-library-o11.scss
+++ b/src/scss/01-foundations/_icon-library-o11.scss
@@ -152,10 +152,6 @@
 	}
 }
 
-.form .dropdown-container:after {
-	top: calc(50% - var(--space-m) / 2);
-}
-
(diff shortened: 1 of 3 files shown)


BenOsodrac:
@os-davidlourenco since the rule is the same for both platforms, we could have this on the form scss file

BenOsodrac (approved):

PR #1185, opened 2026-07-28
ROU-12937: Fix Carousel removeChild NotFoundError on data refresh

gnbm:
### What was happening

On O11 and  ODC, a Carousel with bare `` slides crashed with `NotFoundError: Failed to execute 'removeChild' on 'Node'` (`OS-CLRT-60500`) when a `ParameterOnChange` action re-executed an aggregate feeding the carousel (and an IF around it).

Regressed in **v2.29.0** by commit `c30702391` (PR #1138, ROU-12567): `_prepareCarouselItems()` started reparenting bare `` slides into a newly created wrapper `<div>`:

```ts
if (item.tagName === 'IMG') {
    const wrapper = document.createElement(...Div);
    item.replaceWith(wrapper);   // moves a platform-owned node
    wrapper.appendChild(item);
    wrapper.classList.add(Enum.CssClass.SplideSlide);
}
```

The `` is a **platform-owned** node. On O11 and  ODC, the runtime is React; moving it invalidates React's fiber record of the node's parent, so the next reconcile (triggered by the aggregate refresh) calls `parent.removeChild(img)` against a parent that no longer contains it → throw. Every provider `redraw()` re-ran this and re-wrapped the fresh nodes, making it self-sustaining.

### What was done

Stop reparenting and keep the ARIA roles valid without moving any DOM — 3 changes in `Splide.ts`:

1. **`_prepareCarouselItems()`** — restored the pre-regression class-add loop: every item (incl. ``) just gets the `splide__slide` class; no node is created or moved. Kept the `Array.from` snapshot.
2. **`_applyListRoles()`** — skip list semantics when the list has bare `` slides: a `role="list"` whose children aren't list items is invalid ARIA, and an image can't take `role="listitem"` without losing its image semantics. Also drops a `role="list"` left from a previous pass when content changes across platform refreshes. The images stay exposed to screen readers via their alt text.
3. **`_setListRoles()`** — strip the role Splide assigns to a bare `` slide: `role="presentation"` is valid on an image but removes it from the accessibility tree, so stripping it keeps the image announced via its alt text. Also strip `role="tabpanel"` from slides that **are** native list elements (`ul`/`ol`/`li`), via a `tagName` check.

PR #1138's other improvements (WCAG pagination target size, list/listitem roles for non-img slides, root `role="presentation"`) are preserved untouched.

### Test Steps

1. Open the sample page 
2. Go to the "Test Carousel OnParameterChanged" section.
3. Toggle the `Test_Carousel_OnParameterChanged` switch a few times (re-runs `OnParametersChanged`).
4. Before fix: `NotFoundError … removeChild` in the console. After fix: no error; carousel keeps rendering.

### Screenshots

N.A.

### Checklist

-   [x] tested locally
-   [x] documented the code

### DIFF
--- a/src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
+++ b/src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
@@ -142,8 +142,8 @@ namespace Providers.OSUI.Carousel.Splide {
 				for (const item of _childrenList) {
 					if (!item.classList.contains(Enum.CssClass.SplideSlide)) {
 						// Never create or move platform-owned nodes here (ROU-12937): reparenting an <img>
-						// (e.g. wrapping it in a <div>) invalidates React's fiber bookkeeping on ODC and
-						// crashes the next reconcile with NotFoundError on removeChild. Only mutate classes;
+						// (e.g. wrapping it in a <div>) invalidates React's fiber bookkeeping and crashes
+						// the next reconcile with NotFoundError on removeChild. Only mutate classes;
 						// invalid ARIA roles on <img> slides are handled non-destructively in _setListRoles.
 						item.classList.add(Enum.CssClass.SplideSlide);
 					}
@@ -192,22 +192,33 @@ namespace Providers.OSUI.Carousel.Splide {
 		private _setListRoles(): void {
 			// Remove role="tabpanel" from slides that contain img, ul, ol, or li — elements for
 			// which tabpanel ownership is invalid or creates conflicting semantics
-			this.selfElement.querySelectorAll(OSFramework.OSUI.Constants.Dot + Enum.CssClass.SplideSlide).forEach((slide) => {
-				const _slideEl = slide as HTMLElement;
-				// A bare <img> slide (no wrapper — see ROU-12937) can hold no valid list/tab role, so
-				// strip whatever role Splide assigned to it (e.g. presentation/tabpanel). Mirrors the
-				// tabpanel strip below for slides that merely contain media — attribute-only, no reparenting.
-				if (_slideEl.tagName === 'IMG') {
-					OSFramework.OSUI.Helper.Dom.Attribute.Remove(_slideEl, OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName);
-					return;
-				}
-				if (
-					OSFramework.OSUI.Helper.Dom.Attribute.Get(_slideEl, OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName) === OSFramework.OSUI.Constants.A11YAttributes.Role.TabPanel &&
-					_slideEl.querySelector('img, ul, ol, li')
-				) {
-					OSFramework.OSUI.Helper.Dom.Attribute.Remove(_slideEl, OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName);
-				}
-			});
+			this.selfElement
+				.querySelectorAll(OSFramework.OSUI.Constants.Dot + Enum.CssClass.SplideSlide)
+				.forEach((slide) => {
+					const _slideEl = slide as HTMLElement;
+					// A bare <img> slide (no wrapper — see ROU-12937) can hold no valid list/tab role, so
+					// strip whatever role Splide assigned to it (e.g. presentation/tabpanel). Mirrors the
+					// tabpanel strip below for slides that merely contain media — attribute-only, no reparenting.
+					if (_slideEl.tagName === 'IMG') {
+						OSFramework.OSUI.Helper.Dom.Attribute.Remove(
+							_slideEl,
+							OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName
+						);
+						return;
+					}
+					if (
+						OSFramework.OSUI.Helper.Dom.Attribute.Get(
+							_slideEl,
+							OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName
+						) === OSFramework.OSUI.Constants.A11YAttributes.Role.TabPanel &&
+						_slideEl.querySelector('img, ul, ol, li')
+					) {
+						OSFramework.OSUI.Helper.Dom.Attribute.Remove(
+							_slideEl,
+							OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName
+						);
+					}
+				});
 
 			if (this._hasList && this._carouselListWidgetElem) {
 				// Dynamic content: poll until the List widget finishes loading before applying roles
--- a/src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
+++ b/src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
@@ -43,13 +43,10 @@ namespace Providers.OSUI.Carousel.Splide {
 		// aren't list items is invalid — the images stay exposed to screen readers via alt text.
 		private _applyListRoles(listEl: HTMLElement): void {
 			const _isNativeList = listEl.tagName === 'UL' || listEl.tagName === 'OL';
-			// The direct children are the slides that would receive the list item role. This check
-			// must stay direct-child-only (a wrapped image, div > img, must still get list roles),
-			// which the Dom helpers can't express — ClassSelector matches by class, TagSelector
-			// matches any descendant — so read the direct children and check them by tag.
-			const _slides = Array.from(listEl.children);
-			const _hasNativeListChildren = _slides.some((slide) => ['LI', 'OL', 'UL'].includes(slide.tagName));
-			const _hasImageSlides = _slides.some((slide) => slide.tagName === 'IMG');
+			const _hasNativeListChildren = listEl.querySelector(':scope > li, :scope > ul, :scope > ol') !== null;
+			// Direct-child <img> slides only: a wrapped image (div > img) must still get list roles,
+			// so this can't use ClassSelector (class-based) or TagSelector (matches any descendant).
+			const _hasImageSlides = listEl.querySelector(':scope > img') !== null;
 
 			if (_isNativeList || _hasNativeListChildren || _hasImageSlides) {
 				// Content can change across platform refreshes (e.g. List widget data) — drop a list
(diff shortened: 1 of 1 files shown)


src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -48,7 +48,13 @@ namespace Providers.OSUI.Carousel.Splide {
 
 			listEl.setAttribute('role', 'list');
```
OS-susmitabhowmik: We could add a check to set the list role here only if the carousel does not contain images. Since we are not adding the `listitem` role for images, in the case of a carousel with only images this list wouldn't contain any list items

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -48,7 +48,13 @@ namespace Providers.OSUI.Carousel.Splide {
...
 			listEl.querySelectorAll(':scope > *').forEach((item) => {
-				item.setAttribute('role', 'listitem');
+				// Never assign a list role to a bare <img> slide: neither role="listitem" nor Splide's
+				// role="presentation" is valid on an image. Strip any assigned role instead (ROU-12937).
```
OS-susmitabhowmik: I think think I added this comment previously but it is misleading 😅 `role="presentation"` is valid on an image but it would remove the image from being read by a screen reader

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -192,15 +192,33 @@ namespace Providers.OSUI.Carousel.Splide {
...
+							OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName
+						);
+						return;
+					}
+					if (
+						OSFramework.OSUI.Helper.Dom.Attribute.Get(
+							_slideEl,
+							OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName
+						) === OSFramework.OSUI.Constants.A11YAttributes.Role.TabPanel &&
+						_slideEl.querySelector('img, ul, ol, li')
```
OS-susmitabhowmik:
```suggestion
						(['UL', 'OL', 'LI'].includes(_slideEl.tagName))
```
We could remove images from this query selector because we are removing all roles from images in the previous if statement. We could also update the query to match the code above and validate the `tagName` to remove the `tabpanel` role from matching elements

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
gnbm: Applied in 6d4dfa019 — `_applyListRoles` now skips list semantics entirely when the list has bare `img` direct children, so no `role="list"` is exposed without list items. It also drops a `role="list"` left from a previous pass, for the case where List-widget content changes across platform refreshes. The images stay in the accessibility tree and are announced via their alt text.

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
gnbm: Reworded in 6d4dfa019 — the comment now states that `role="presentation"` is valid on an image but removes it from the accessibility tree, hiding it from screen readers, and that stripping the role keeps the image announced via its alt text. Behavior unchanged, justification corrected.

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
gnbm: Applied the suggestion in 6d4dfa019. It also fixes a latent gap: a slide that *is* a bare `li`/`ul`/`ol` didn't match the old descendant `querySelector` (it only searches descendants), so its `tabpanel` survived in exactly the native-list case where `_applyListRoles` early-returns and nothing overrides it. Slides that merely contain media get `listitem` set right after, which already overwrote `tabpanel`. One accepted edge: a mixed carousel (bare images + container slides holding media) now skips list roles, so a container slide there can keep `tabpanel` — contrived shape, no worse than pre-2.29.

OS-susmitabhowmik (dismissed):

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -37,12 +37,21 @@ namespace Providers.OSUI.Carousel.Splide {
...
 		private _applyListRoles(listEl: HTMLElement): void {
 			const _isNativeList = listEl.tagName === 'UL' || listEl.tagName === 'OL';
 			const _hasNativeListChildren = listEl.querySelector(':scope > li, :scope > ul, :scope > ol') !== null;
+			const _hasImageSlides = listEl.querySelector(':scope > img') !== null;
 
-			if (_isNativeList || _hasNativeListChildren) {
+			if (_isNativeList || _hasNativeListChildren || _hasImageSlides) {
+				// Content can change across platform refreshes (e.g. List widget data) — drop a list
+				// role applied on a previous pass so it never wraps children that aren't list items
+				if (_hasImageSlides && listEl.getAttribute('role') === 'list') {
```
joselrio: `OSFramework.OSUI.Helper.Dom.Attribute.Get` instead `.getAttribute`

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -37,12 +37,21 @@ namespace Providers.OSUI.Carousel.Splide {
...
+					listEl.removeAttribute('role');
```
joselrio: `OSFramework.OSUI.Helper.Dom.Attribute.Remove` instead?

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -37,12 +37,21 @@ namespace Providers.OSUI.Carousel.Splide {
...
 		// Skips native list elements (ul/ol) and elements whose direct children are native list
-		// elements (li/ul/ol), since those already carry implicit list semantics.
+		// elements (li/ul/ol), since those already carry implicit list semantics. Also skips lists
+		// with bare <img> slides: images can't take role="listitem" without replacing their image
+		// semantics, wrapping them is off-limits, and a role="list" whose children
+		// aren't list items is invalid — the images stay exposed to screen readers via alt text.
 		private _applyListRoles(listEl: HTMLElement): void {
 			const _isNativeList = listEl.tagName === 'UL' || listEl.tagName === 'OL';
 			const _hasNativeListChildren = listEl.querySelector(':scope > li, :scope > ul, :scope > ol') !== null;
+			const _hasImageSlides = listEl.querySelector(':scope > img') !== null;
```
joselrio: `OSFramework.OSUI.Helper.Dom.ClassSelector` instead?

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
joselrio: `OSFramework.OSUI.Helper.Dom.Attribute.Set`, 
`OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName`,
`OSFramework.OSUI.Constants.A11YAttributes.Role.Listitem`

joselrio: `OSFramework.OSUI.Helper.Dom.Attribute.Set`,
`OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName`,
`OSFramework.OSUI.Constants.A11YAttributes.Role.List`

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -135,17 +144,11 @@ namespace Providers.OSUI.Carousel.Splide {
...
-							wrapper.appendChild(item);
-							wrapper.classList.add(Enum.CssClass.SplideSlide);
-						} else {
-							item.classList.add(Enum.CssClass.SplideSlide);
-						}
+						// Never create or move platform-owned nodes here: reparenting an <img>
+						// (e.g. wrapping it in a <div>) invalidates React's fiber bookkeeping and crashes
+						// the next reconcile with NotFoundError on removeChild. Only mutate classes;
+						// ARIA roles on <img> slides are stripped non-destructively in _setListRoles.
+						item.classList.add(Enum.CssClass.SplideSlide);
```
joselrio: `OSFramework.OSUI.Helper.Dom.Styles.AddClass`

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -135,17 +144,11 @@ namespace Providers.OSUI.Carousel.Splide {
 				// Add the placeholder content already with the correct html structure per item, expected by the library
 				for (const item of _childrenList) {
 					if (!item.classList.contains(Enum.CssClass.SplideSlide)) {
```
joselrio: `OSFramework.OSUI.Helper.Dom.Styles.ContainsClass`

joselrio (changes requested):

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
```diff
@@ -190,17 +193,36 @@ namespace Providers.OSUI.Carousel.Splide {
...
-					OSFramework.OSUI.Helper.Dom.Attribute.Get(_slideEl, OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName) === OSFramework.OSUI.Constants.A11YAttributes.Role.TabPanel &&
-					_slideEl.querySelector('img, ul, ol, li')
-				) {
-					OSFramework.OSUI.Helper.Dom.Attribute.Remove(_slideEl, OSFramework.OSUI.Constants.A11YAttributes.Role.AttrName);
-				}
-			});
+			// Remove role="tabpanel" from slides that are native list elements (ul/ol/li) — the same
+			// cases where _applyListRoles skips custom roles, so nothing overrides the conflicting
+			// tabpanel afterwards
+			this.selfElement
```
joselrio, line 224: I’ve raised this before, but this is exactly the type of approach that, when properly evaluated, may not be ideal to include at the component level.

This kind of logic should instead be documented and left for developers to implement based on their specific use cases. With the current approach, every time the Carousel is mounted, we iterate through all carousel items, which will inevitably have a performance impact — especially as the number of items grows.

Additionally, this does not fully address the accessibility (A11Y) concern. There is still required work on the developer side to ensure accessibility compliance, meaning this PR alone will not prevent the issue it aims to solve.

Raising this again for visibility, as we should carefully consider whether this is the right place and approach for this kind of behaviour.

src/scripts/Providers/OSUI/Carousel/Splide/Splide.ts
gnbm: You're right. I'd assumed `TagSelector` was tag-only, but it just wraps `querySelector` and is used across the codebase with compound selectors (e.g. `TagSelector(selfElement, 'input:checked')`, `TagSelector(document.head, '[name="viewport"]')`). So this now goes through the helper:

```ts
const _hasImageSlides = OSFramework.OSUI.Helper.Dom.TagSelector(listEl, ':scope > img') !== undefined;
```

The `:scope >` keeps it direct-child-only (a wrapped image, `div > img`, must still receive list roles). I used `TagSelector` rather than `ClassSelector` only because the check is by tag/position, not class. Also converted the sibling `_hasNativeListChildren` and the `:scope > *` loop to `TagSelector`/`TagSelectorAll`, so `_applyListRoles` no longer has any raw `querySelector`.

joselrio (approved):

PR #1190, opened 2026-07-31
ROU-12875: Align interaction patterns with Figma

OS-giulianasilva:
This PR is for aligning the interaction patterns and their form-control states with the Figma design specs.

### What was happening
* Action Sheet used elevation 2 and inherited generic button sizing, so its options did not match the spec.
* Animated Label had no hover state, no focus ring and no disabled treatment. On focus the floating label turned primary, and the error state coloured both the border and the label.
* Carousel disabled arrows were faded to 30% opacity, dimming the icon itself, and the active pagination dot stretched in width so dots had inconsistent sizes.
* Date Picker calendar cells, month/year controls and the time input used hardcoded sizes and radii instead of tokens.
* The Dropdown, Dropdown Search, Dropdown Server Side and Dropdown Tags arrow was sized off font-size, so its box and position drifted between variants.
* Button Group items were 48px tall while the Button widget is 40px, so a group placed next to a button did not line up.
* The hover border on Input, Textarea, Checkbox, Radio Button and Switch resolved to the input press tier instead of the input hover tier. The Checkbox hover rule re-declared its resting colour, so hovering changed nothing.
* The Radio Button resting border used the generic border role, which resolves to the same value as the input hover tier, leaving its hover state invisible.
* The [class*='background-'] button rule forced inverse text colour, making label text invisible on background-white buttons.

### What was done
* Action Sheet: raised to elevation 3, added a colour knob, and routed its buttons through the button CSS API with 44px height and small action typography.
* Animated Label: added hover, focus-ring, disabled and error knobs to the component CSS API. The label keeps its subtlest colour on focus and only the container border reacts to the error state.
* Carousel: replaced the opacity fade on disabled arrows with a disabled overlay that keeps the icon fully visible, and made all pagination dots a uniform 12px so only colour changes on the active dot.
* Date Picker: migrated the Flatpickr override sheet to tokens for cell sizes, hover surfaces, radii and the time input.
* Dropdown family: gave the arrow an explicit size box and colour knob so it renders consistently across the widget, Search, Server Side and Tags variants.
* Button Group: added a height knob defaulting to 40px, applied as both height and max-height, and dropped the responsive override that pushed it to 48px.
* Form controls: pointed the hover border of Input, Textarea, Checkbox, Radio Button, Switch and Animated Label at the input hover tier, and aligned the Radio Button resting border with the other controls so its hover step shows.
* Dark theme: added the missing input hover tier override so the new hover borders do not fall back to the light value in dark mode.
* Button: excluded background-white from the inverse-text utility rule.

### Test Steps
1. Run the dev server for ODC and open the Patterns sample screen.
2. Hover Input, Textarea, Checkbox, Radio Button and Switch and confirm each border changes to the input hover colour, one tier lighter than press.
3. Confirm a checked Checkbox or Radio Button keeps its primary border while hovered, and a disabled one does not react at all.
4. Place a Button Group next to a Button and confirm both are 40px tall and their top and bottom edges line up.
5. Hover the Animated Label container, then focus it, then disable it, then trigger a validation error, confirming each state matches the spec.
6. Move a Carousel to its first slide and confirm the previous arrow shows a disabled overlay with the icon still visible and does not respond to clicks.
7. Confirm all Carousel pagination dots are the same size and only the active dot changes colour.
8. Open the Date Picker and Dropdown variants and confirm cell sizing, radii and the arrow glyph are consistent.
9. Add theme-dark to the screen root and repeat steps 2 and 5, confirming the hover borders stay dark-appropriate.

### Screenshots
(prefer animated gif)

### Checklist
* [ ] tested locally
* [ ] documented the code

### DIFF
--- a/src/scss/03-widgets/_button-group.scss
+++ b/src/scss/03-widgets/_button-group.scss
@@ -16,8 +16,9 @@
 		--osui-button-group-color: var(--color-text-subtle);
 		--osui-button-group-disabled-overlay: #{$token-state-disabled};
 		// Matches the default .btn height so a button group sits flush with
-		// buttons placed alongside it
-		--osui-button-group-height: #{$token-scale-1000};
+		// buttons placed alongside it, while still growing with larger font
+		// sizes and accessibility settings
+		--osui-button-group-min-height: #{$token-scale-1000};
 		// ───────────────────────────────────────────────────────────────────
 
 		background-color: var(--osui-button-group-background);
@@ -27,8 +28,7 @@
 		cursor: pointer;
 		font-size: $token-font-size-400;
 		font-weight: $token-font-weight-medium;
-		height: var(--osui-button-group-height);
-		max-height: var(--osui-button-group-height);
+		min-height: var(--osui-button-group-min-height);
 		padding: $token-scale-0 $token-scale-400;
 		position: relative;
 		white-space: nowrap;
--- a/src/scss/03-widgets/_button-group.scss
+++ b/src/scss/03-widgets/_button-group.scss
@@ -15,6 +15,9 @@
 		--osui-button-group-border-color: var(--color-border);
 		--osui-button-group-color: var(--color-text-subtle);
 		--osui-button-group-disabled-overlay: #{$token-state-disabled};
+		// Matches the default .btn height so a button group sits flush with
+		// buttons placed alongside it
+		--osui-button-group-height: #{$token-scale-1000};
 		// ───────────────────────────────────────────────────────────────────
 
 		background-color: var(--osui-button-group-background);
@@ -24,7 +27,8 @@
 		cursor: pointer;
 		font-size: $token-font-size-400;
 		font-weight: $token-font-weight-medium;
-		height: $token-scale-1200;
+		height: var(--osui-button-group-height);
+		max-height: var(--osui-button-group-height);
 		padding: $token-scale-0 $token-scale-400;
 		position: relative;
 		white-space: nowrap;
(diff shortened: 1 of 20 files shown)


src/scss/03-widgets/_button-group.scss
```diff
@@ -24,7 +27,8 @@
 		cursor: pointer;
 		font-size: $token-font-size-400;
 		font-weight: $token-font-weight-medium;
-		height: $token-scale-1200;
+		height: var(--osui-button-group-height);
+		max-height: var(--osui-button-group-height);
```
BenOsodrac: @OS-giulianasilva maybe a better solution would be to set only min-height (and rename the var as well), something like this: `min-height: var(--osui-button-group-min-height);`, this way it will scale to different font sizes and a11y settings. 

We also might need to review this on other patterns for a similar approach (not for this PR)

src/scss/04-patterns/03-interaction/animated-label/_animated-label.scss
```diff
@@ -147,29 +152,33 @@
...
 	}
 
+	// Hover — subtle border emphasis
+	&:hover:not(:focus-within):not(:has([disabled])) {
+		border-color: var(--osui-animated-label-hover-border-color);
+	}
+
 	&:focus-within {
 		border-color: var(--osui-animated-label-focus-border-color);
+		box-shadow: 0 0 0 $token-border-size-050 var(--osui-animated-label-focus-ring-color);
```
BenOsodrac: I would say maybe this deserves a css variable on the component API for the box-shadow, something like `--osui-animated-label-focus-shadow`

BenOsodrac (changes requested):

BenOsodrac (approved):

PR #1192, opened 2026-07-31
ROU-12877: Align number patterns with figma

JoaoFerreira-FrontEnd:
This PR is for aligning rating pattern with the Figma design.

### What was happening

- The primary color was not being settled. 

### What was done
- fix rating icons color;

### Test Steps

1. Open the screen of each pattern. Confirm the pattern matches Figma.

### Checklist

-   [x] tested locally
-   [ ] documented the code

### DIFF
--- a/src/scss/04-patterns/05-numbers/rating/_rating.scss
+++ b/src/scss/04-patterns/05-numbers/rating/_rating.scss
@@ -11,7 +11,7 @@
 	// ───────────────────────────────────────────────────────────────────
 
 	// Internal colour vars — overrideable per-instance
-	--token-rating-filled-color: #{$token-icon-primary};
+	--token-rating-filled-color: var(--color-primary};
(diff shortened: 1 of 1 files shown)


src/scss/04-patterns/05-numbers/rating/_rating.scss
```diff
@@ -11,12 +11,8 @@
 	// ───────────────────────────────────────────────────────────────────
 
 	// Internal colour vars — overrideable per-instance
-	--token-rating-filled-color: #{$token-semantics-warning-base};
-	--token-rating-empty-color: color-mix(
-		in srgb,
-		var(--token-rating-filled-color) 38%,
-		#{$token-primitives-base-white}
-	);
+	--token-rating-filled-color: #{$token-icon-primary};
```
OS-giulianasilva: I think we can use this variable instead of the token directly.
```suggestion
	--token-rating-filled-color: var(--color-primary};
```

PR #1207, opened 2026-08-11
ROU-12946: Let SanitizeDropdownValues own the option text, and keep the icon and image inside their attributes

gnbm:
This PR is for keeping the option icon and image inside the attributes OSUI builds for them, and
for letting `SanitizeDropdownValues` decide whether option text is sanitized.

[Sample page](url)

### What was happening

**The icon and image prefixes.** `_getOptionIconPrefix()` and `_getOptionImagePrefix()` build the
prefix prepended to each option label, and their return value reaches the dropbox `innerHTML`
through our `labelRenderer`, which the library writes out without escaping. Both interpolated
`image_url_or_class` into an attribute of a hand built String:

-   the icon put it in a `class` attribute with no escaping at all, so `">`
    injected live markup;
-   the image relied on `Helper.Sanitize()`, which replaces angle brackets but not the double quote,
    and `URL.IsImage()` accepts an image extension followed by a query string, so
    `photo.png?" onerror="...` closed the `src` attribute early.

Neither depended on `SanitizeDropdownValues`: `image_url_or_class` is OSUI's own field and never
reaches the library.

**The option text.** `getProviderConfig()` ran `Helper.Sanitize()` over `label`, `description` and
`group_name` unconditionally, which put OSUI on the wrong side of `SanitizeDropdownValues` in both
of its positions:

-   with it **True**, `prepareOption()` already runs every label — group titles included — and every
    description through `secureText()`, and escapes them again at each attribute boundary
    (`aria-label`, `data-tooltip`, `data-value`). Our pass was redundant.
-   with it **False**, the app has asked for its markup to render, and our pass was the only thing
    stopping it.

Ours was also the lossier of the two. The `<` → `‹` substitution is never undone, so it reached the
app back through `getSelectedValues()`, was announced as a guillemet by assistive technology, and
made the text unsearchable — the library escapes to entities, derives `labelNormalized` from the raw
text, and decodes the entities back for accessible names.

### What was done

-   Both prefixes are built through the DOM and serialized (`createElement` + `outerHTML`) instead
    of interpolated into a String, so the browser escapes the attribute values. The image `src` no
    longer goes through `Sanitize()`, which was both insufficient here and lossy for urls carrying
    angle brackets.
-   Removed the `Helper.Sanitize()` pass over `label`, `description` and `group_name`, so
    `SanitizeDropdownValues` decides. `group_name` untouched also means two group names differing
    only by their angle brackets no longer collapse into one group.
-   Added `Icon`/`Image` to `GlobalEnum.HTMLElement` and `Src` to `GlobalEnum.HTMLAttributes`.
-   Recorded in ADR-0003, superseding section B of ADR-0002.

Resulting behaviour, verified against 1.4.0's escaping:

| `SanitizeDropdownValues` | markup in option text | `5" screen` / `R&D` | search | `getSelectedValues()` |
|---|---|---|---|---|
| **False** (default) | renders, as today | display correctly | matches | values as provided |
| **True** | escaped, shown literally, never executes | display correctly | matches the authored text | values as provided |

No API, property or markup change.

### Test Steps

1. `npm run build`
2. Serve `dist/` and load a page creating a `search`-mode `virtual-select` Dropdown with grouped
   options carrying `image_url_or_class: '">'` and
   `image_url_or_class: 'x.png?" onerror="window.__pwned=1'`, plus two groups named `<Legacy>` and
   `‹Legacy›`
3. Open the dropbox and inspect `.vscomp-option`, with `SanitizeDropdownValues` False and True

Expected in both settings: no extra element inside the option, no `onerror` on the image and
`window.__pwned` undefined; legitimate values unchanged (`fa fa-user` yields the same class list,
`/img/photo.png?w=1&h=2` reads back identically from `getAttribute('src')`); `<Legacy>` and
`‹Legacy›` remain two groups. With False, `<b>Region</b>` in a label or group name renders bold as
it does today; with True it is shown literally and never executes.

Verified in the browser for each payload, and both flag states were run through 1.4.0's
`secureText()`, `getAriaLabelText()` and `getAttributesText()` to confirm the table above.
`tsc --noEmit`, `npm run lint` and `prettier --check` are clean.

### Screenshots

n/a — the observable difference is DOM/script execution rather than visual.

### Checklist

-   [x] tested locally
-   [x] documented the code


gnbm:
Closing this without merging.

Re-reviewing the original findings against VirtualSelect **1.4.0** rather than the 1.1.0 in `node_modules` shows the library's own 1.4.0 hardening already covers them. `v1.2.0` and `v1.3.0` contain none of it; 1.4.0 introduced `getAriaLabelText()`, `getAttributesText()` and `escapeAttributeValue()`, which escape:

- `customData` -> `aria-label` (`prepareOption` + the aria-label boundary)
- `label` -> `aria-label` and `data-tooltip`, unconditionally
- `data-value`

So the upgrade in #1201 is what closed those, and no OSUI-side change is needed for them.

**One finding is not covered by any library version and remains open**, tracked back to ROU-12946:

`_getOptionIconPrefix()` and `_getOptionImagePrefix()` interpolate `image_url_or_class` into a `class` and a `src` attribute of a hand built String, and that String reaches the dropbox `innerHTML` through our `labelRenderer`. `image_url_or_class` is OSUI's own field and never reaches the library, so `SanitizeDropdownValues` and `enableSecureText` have never applied to it. `'">'` takes the icon branch and injects live markup; `'photo.png?" onerror="...'` passes `URL.IsImage()` and closes the `src` attribute early. Building both prefixes through the DOM and serializing them fixes it with no change for legitimate values.

Severity is moderate: that field usually holds a developer-authored icon class rather than user data.

The work is preserved at commit `c3e4ebead` on this branch if it is wanted later.

PR #1212, opened 2026-08-13
ROU-12925: Align navigation patterns with Figma

OS-giulianasilva:
This PR is for aligning the four navigation patterns — Submenu, Tabs, Timeline and Wizard — with their Figma specs.

### What was happening
* Submenu items had no pressed state and no disabled state at all, and the panel used a 4px radius instead of the 12px in Figma.
* Tab header items rounded all four corners, so the item lifted off the header's bottom border.
* Timeline spacing came from the icon line's bottom margin rather than the item, and the icon line kept a bottom margin that Figma does not have.
* The Wizard connector was positioned from the marker's edge. A 2px line meeting a round marker at a tangent gets notched by the curve, so the track read as broken at every step. With the empty (dot) marker in vertical orientation the line also fell short at both ends.
* Item labels across all four patterns inherited a line height instead of the 21px Figma specifies.

### What was done
* Submenu: added a pressed state (`bg/neutral/subtlest/press`) and a CSS-only disabled state (`.is-disabled`), moved the panel to `border-radius/300`, and set the item line height. Every context rule that paints an item now carries `:not(:active)` so nothing masks the pressed background.
* Tabs: the header item radius keeps only the top corners, exposed as `--osui-tabs-header-item-border-radius`, plus the item line height. The story now disables one header item through `TabsHeaderItemAPI.DisableTabItem`.
* Timeline: line height and an 8px item margin (wired to the `--osui-timeline-gap` knob that was declared but never consumed), and the icon line's bottom margin removed. New `NoIcon` story for the Figma `Type=No icon` variant.
* Wizard: the connector now runs marker centre to marker centre and passes behind each marker, which sits on a higher layer and covers the overlap, so the joint reads continuous for every marker type and no longer depends on `--osui-wizard-icon-size`. Also 8px icon padding, `bg/neutral/subtlest/default` background, label line height, 14px medium number, 20px glyph and rounded connector ends. Two stories added for the icon and empty markers.
* This branch was merged with its base. The base already carried the content-pattern work this branch duplicated, so those files take the base version. The Wizard was rewritten upstream as a TypeScript pattern and its SCSS moved to `04-patterns/02-content/wizard/`, so the old `04-navigation/_wizard.scss` is dropped and every change this branch made to it is ported onto the new files and markup contract.

Note on tokens: 21px has no token — the `font-line-height` scale steps from 20px to 24px — so each pattern declares the literal on its own `--osui-*-line-height` knob with a TODO, matching what the base already does for the ListItemContent, ChatMessage and Section knobs. Worth raising upstream now that there are several call sites.

### Test Steps
1. Run `npm run dev -- --target ODC` and open Storybook.
2. Submenu: open the panel, press and hold an item — the background is `#f2f2f2` and does not revert to the hover colour. The last item is disabled: muted text, no hover or pressed feedback. The panel corners are 12px.
3. Tabs: the header items round only at the top. The `Settings` tab is disabled — muted, no hover background, and clicking it does not change the pane.
4. Timeline: check the `Default` and `NoIcon` stories. In `NoIcon` the 24px circle collapses to an 8px dot.
5. Wizard: check `Wizard`, `Marker: icon` and `Marker: empty (dot)`, each in horizontal and vertical. Zoom into where the connector meets a marker — there should be no notch or gap at any joint, for any marker size.
6. Toggle `Item.ReverseLabelPosition` and confirm the label flips side in both orientations.
7. Run `npm run build` — it must finish with no errors and a clean lint.

### Screenshots
(prefer animated gif)

### Checklist
* [x] tested locally
* [x] documented the code
* [x] clean all warnings and errors of eslint


src/scss/04-patterns/02-content/wizard/_wizard-item.scss
```diff
@@ -10,18 +10,28 @@
...
 	--osui-wizard-icon-border-color: var(--color-border-input);
 	--osui-wizard-icon-border-width: #{variables.$token-border-size-050};
 	--osui-wizard-icon-color: #{variables.$token-text-subtlest};
+	--osui-wizard-icon-font-weight: #{variables.$token-font-weight-medium};
+	--osui-wizard-icon-glyph-size: #{variables.$token-font-size-500};
+	--osui-wizard-icon-padding: #{variables.$token-space-200};
 	--osui-wizard-icon-size: #{variables.$token-scale-1000};
 	--osui-wizard-label-color: var(--color-text);
+	// TODO: needs a line-height token — the font-line-height scale has no 21px step (Figma Wizard label).
+	--osui-wizard-label-line-height: 21px;
```
BenOsodrac: @OS-giulianasilva does the wizard really need such a specific line.-height? I would to align with Jéssica/Mariana and see if we can use the one with the closest value, surely will work just fine

src/scss/04-patterns/04-navigation/submenu/_submenu.scss
```diff
@@ -9,13 +9,20 @@
...
-	--osui-submenu-items-border-color: transparent;
-	--osui-submenu-items-shadow: #{variables.$token-elevation-2};
-	--osui-submenu-item-color: var(--color-text-subtlest);
+	--osui-submenu-items-border-color: var(--color-border);
+	--osui-submenu-items-border-radius: #{variables.$token-border-radius-300};
+	--osui-submenu-items-shadow: #{variables.$token-elevation-3};
+	--osui-submenu-item-color: var(--color-text);
+	--osui-submenu-item-font-size: #{variables.$token-font-size-350};
+	// TODO: needs a line-height token
+	--osui-submenu-item-line-height: 21px;
```
BenOsodrac: Same as Wizard

src/scss/04-patterns/04-navigation/tabs/_tabs.scss
```diff
@@ -9,7 +9,12 @@
 .osui-tabs {
 	// ─── Component CSS API ─────────────────────────────────────────────
 	--osui-tabs-border-color: var(--color-border);
+	--osui-tabs-header-gap: #{variables.$token-space-800}; // horizontal only; Justified stretches instead
 	--osui-tabs-header-item-color: var(--color-text-subtlest);
+	--osui-tabs-header-item-font-size: #{variables.$token-font-size-350};
+	--osui-tabs-header-item-border-radius: #{variables.$token-border-radius-100} #{variables.$token-border-radius-100} 0 0;
+	// TODO: needs a line-height token
+	--osui-tabs-header-item-line-height: 21px;
```
BenOsodrac: same

BenOsodrac (changes requested):
Can we push back to Design the line-height decision?

Also, when we do changes to CSS API vars, ask Claude to update the CSS API Reference page on storybook ;)

src/scss/04-patterns/02-content/wizard/_wizard-item.scss
OS-giulianasilva: I talked with Jéssica and she said that's ok if it's less than 3px of difference. Most of them differed in 1px from the tokens, so I updated them accordingly.

BenOsodrac (approved):

PR #1246, opened 2026-09-02
ROU-13020: Fixed padding for unselected dropdown

os-davidlourenco:
This PR fixes dropdown toggle padding so unselected VirtualSelect dropdowns align correctly in LTR and RTL.

### What was happening
* The VirtualSelect vendor styles applied hardcoded `padding` on `.vscomp-toggle-button`, which conflicted with OSUI token-based spacing.
* The clear-button LTR rule (`padding-right: 54px`) also applied in RTL, where a separate `padding-left` rule is expected.
* Extra `padding-block` on the OSUI toggle button compounded the misalignment for unselected dropdowns.

### What was done
* Removed vendor hardcoded toggle padding from `_virtualselect_lib.scss` (default, RTL, and show-value-as-tags variants).
* Scoped the clear-button padding rule to LTR only: `.has-clear-button:not(.text-direction-rtl)`.
* Removed `padding-block` from the OSUI toggle button in `_virtualselect.scss`.

### Test Steps
1. Open a screen with a Dropdown (VirtualSelect) without a selected value.
2. Verify toggle height and horizontal padding look correct in LTR.
3. Switch the app to RTL and repeat — clear-button spacing should use `padding-left`, not `padding-right`.
4. Test dropdowns with clear button enabled, with tags mode, and with a selected value.
5. Confirm no visual regression for keep-always-open dropdowns.

### Screenshots
(prefer animated gif)

### Checklist
* [ ] tested locally
* [ ] documented the code

### DIFF
--- a/src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
+++ b/src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
@@ -340,7 +340,6 @@
 	height: var(--vscomp-toogle-btn-height);
 	line-height: var(--vscomp-toogle-btn-height);
 	min-width: 180px;
-	padding-block: variables.$token-scale-100;
 	padding-inline-end: variables.$token-scale-1000;
 	padding-inline-start: variables.$token-scale-400;
 	position: relative;
--- a/src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
+++ b/src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
@@ -38,6 +38,16 @@
 				}
 			}
 		}
+
+		.vscomp-wrapper.show-value-as-tags:not(.has-value) .vscomp-toggle-button {
+			padding-block: 0;
+			padding-inline-end: 54px;
+			padding-inline-start: variables.$token-scale-400;
+		}
+
+		.vscomp-wrapper.show-value-as-tags:not(.has-value) .vscomp-value {
+			padding-block-end: 0;
+		}
 	}
 
 	&-ele .vscomp-clear-icon:after,
@@ -95,6 +105,12 @@
 	}
 
 	&:not(.has-value) {
+		.vscomp-toggle-button {
+			padding-block: 0;
+			padding-inline-end: 54px;
+			padding-inline-start: variables.$token-scale-400;
+		}
+
 		.vscomp-value {
 			color: var(--color-text-subtlest);
 			opacity: 1;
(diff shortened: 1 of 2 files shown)


OS-giulianasilva (changes requested):
@os-davidlourenco We should not modify the `_virtualselect_lib.scss` file like this. This file comes from the Virtual Select library and it will be overridden in a next library update.
If that's the case we should always open a PR to the [lib repo](https://github.com/sa-si-dev/virtual-select) and make sure it does not break the behavior of the lib.
Not sure if this is possible in this case, and if not, you should find a way to override the paddings in the `_virtualselect.scss` that is the file we use for the overrides.

os-davidlourenco:
You are absolutely right, completely forgot that! Changed the approach to override on our side!

src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
```diff
@@ -38,6 +38,16 @@
 				}
 			}
 		}
+
+		.vscomp-wrapper.show-value-as-tags:not(.has-value) .vscomp-toggle-button {
+			padding-block: 0;
+			padding-inline-end: 54px;
+			padding-inline-start: variables.$token-scale-400;
```
OS-giulianasilva, line 45: Is this really needed? It appears that the styles added in line 108 already cover this.

src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
```diff
@@ -95,6 +105,12 @@
 	}
 
 	&:not(.has-value) {
+		.vscomp-toggle-button {
+			padding-block: 0;
+			padding-inline-end: 54px;
```
OS-giulianasilva, line 110: What is the purpose of this `padding-inline-end`?

OS-giulianasilva (changes requested):

src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
OS-giulianasilva: Nevermind. I already understood that this is for the RTL mode.

src/scss/04-patterns/03-interaction/dropdown/provider/_virtualselect.scss
os-davidlourenco: Is because of the carrot icon, to be sure that if dev sets a placeholder long enough doesn't get under the carrot icon

OS-giulianasilva (approved):

PR #1267, opened 2026-09-16
ROU-13054: Remove Helper.URL.IsValid

gnbm:
## ROU-13054: Remove `Helper.URL.IsValid`

> **Scope changed after team discussion.** This PR started as a ReDoS fix to the domain regex
> inside `OSFramework.OSUI.Helper.URL.IsValid`. The team decided to **delete the helper outright**,
> since a major version is coming. The hardened regex is no longer part of this PR.

### What was happening

`Helper.URL.IsValid` validated the domain part of a URL with `([a-z\d]([a-z\d-]*[a-z\d])*)` — a
quantified group nested inside another quantified group with overlapping character classes. On a
non-matching input such as `'a'.repeat(n) + '!'` the engine explored exponentially many partitions
before failing, blocking the browser thread:

| input length | time |
|---|---|
| 25 | 58 ms |
| 30 | 3.7 s (ticket) |
| 35 | > 20 s (ticket) |

### What was done

The method is gone. `Helper.URL` now exposes only `GetFileTypeFromURL` and `IsImage`, both
untouched. Nothing else in the repository changed.
Being a breaking-change also incremented the version to v3.0.0

### Why removal instead of the hardened regex

1. **Zero call sites.** A search across `src/`, `stories/`, `docs-internal/` and `classic-theme/`
   found no callers. The method was dead code for its whole lifetime.
2. **Never public.** It lives in the internal `OSFramework.OSUI.*` namespace, is absent from
   `OutSystems/OSUI/**`, and is not in the deprecated `osui.ts` shim. No Service Studio client
   action or public API signature refers to it.
3. **A major version is the moment to do it.** The previous revision of this PR already marked the
   method `@deprecated` and pointed at the native `URL` constructor. Deleting it now avoids
   shipping a deprecation we would have to carry and then remove later anyway.
4. **The whole scanner conflict disappears.** Three rules interacted with that regex and no single
   form of the code satisfied all of them at once — SonarCloud S7780 on the escaped backslashes,
   SonarCloud S5843 on the regex complexity (which needed an inline `// NOSONAR`, the first in this
   repo), and Wiz 00003 or 00094 depending on whether the pattern was a literal. With the pattern
   deleted, all of them are moot and the `NOSONAR` suppression is withdrawn.

### Replacement for external callers

There is no drop-in replacement and none is added. Any code outside the library that called this
should use the platform's own validation or the native constructor:

```js
try {
    new URL(url);   // throws on invalid input, linear time
} catch {
    // not a valid absolute URL
}
```

Note the native constructor requires an absolute URL, so it rejects the relative-path inputs that
`IsValid` accepted through its `window.location.host + url` fallback. Pass a base as the second
argument if relative input has to be supported.

### Test steps

1. `npm run build` — both targets transpile, and the `lintfix` plus `lint` stages that follow report
   zero errors and zero warnings.
2. Confirm the compiled bundles no longer carry the symbol:
   ```bash
   grep -o "URL\.IsValid" dist/O11.OutSystemsUI.js dist/ODC.OutSystemsUI.js   # no matches
   ```
   `URL.GetFileTypeFromURL` and `URL.IsImage` are still present in both.
3. In a page loading either bundle, `OSFramework.OSUI.Helper.URL.IsValid` is now `undefined`.

### Notes for reviewers

- **No E2E runs for this PR.** The change touches only `src/scripts/OSFramework/OSUI/Helper/`,
  which is outside the four path-derived Cucumber tag folders in `pipelines/pr-pipeline.yaml`.
  Verification was the build, the lint gates and the bundle grep above.
- **Breaking change for anyone calling it directly**, which is why it is scoped to the major
  version. Inside the library the impact is nil.
- The three earlier commits on this branch contain the ReDoS analysis and the hardened regex. They
  are kept for the audit trail of the vulnerability; the final state of the branch removes the code
  they fixed. Squash on merge if you would rather the history read as a single deletion.
- The `bug` label is kept: the ticket is a reported vulnerability and this PR is what resolves it.
  Happy to switch it to `chore` if the team reads a dead-code removal that way instead.

### Checklist

- [x] Built locally, both targets
- [x] ESLint / Prettier clean
- [x] Confirmed no call sites before removal
- [ ] Requires OutSystems module changes — no
- [ ] Requires sample page changes — no


OS-giulianasilva (dismissed):
🔥

OS-giulianasilva (dismissed):

OS-giulianasilva (approved):

